PR #1, opened 2017-03-11
Add build starting point

jaredkhan:
This adds the basic Swift structure and a convenient task runner to get started very quickly and easily. You will find Stylus processing, live reloading, and appropriate console output in this PR.


Package.swift
```diff
@@ -0,0 +1,10 @@
+import PackageDescription
+
+let package = Package(
+  name: "HackersAtCambridgeWebsite",
+  dependencies: [
+      .Package(url: "https://github.com/IBM-Swift/Kitura.git", majorVersion: 1, minor: 6),
```
Pinpickle: Is this non-uniform indentation a Swiftism? This line is 4 spaces from the previous level, but the previous line is 2.

jaredkhan: Nope, just a good old-fashioned mistake

Sources/main.swift
```diff
@@ -0,0 +1,76 @@
+import Foundation
+import Kitura
+
+import KituraStencil
+
+import LoggerAPI
+import HeliumLogger
+
+// Create a new router
```
Pinpickle: All of the comments in the file describe exactly what the next line does (except maybe the logger comments). What's the reason for these comments? 

Comments that explain what code does means we have multiple sources of truth (the comment and the line itself), which gives us more of a maintenance burden.

Is it to show an example for those who aren't familiar with the framework/Swift? Would the documentation for these frameworks be a better source of that?

gulpfile.js
```diff
@@ -0,0 +1,123 @@
+const gulp = require('gulp')
```
Pinpickle: This gulpfile is sick nasty, great work putting it together

README.md
```diff
@@ -0,0 +1,30 @@
+# Hackers at Cambridge Website
+
+This website is written using the Swift web framework [Kitura](https://github.com/IBM-Swift/Kitura) to allow us the flexibility to do what we please with the website and to allow us to try out Swift as an option for server-side programming.
+
+## Installing and running the website
+This project uses the [Gulp task runner](http://gulpjs.com) to make this super simple.
```
Pinpickle: Documentation at the same time as features 🎉

README.md
```diff
@@ -0,0 +1,30 @@
...
+Clone this repository
+
+```
+git clone https://github.com/hackersatcambridge/hac-website.git
+```
+
+And then run:
+
+```
+npm install
```
Pinpickle: Is there a reason we're using npm instead of Yarn? Not a blocker as it is very easy to switch later.

Sources/main.swift
jaredkhan: You're right, many of these are superfluous.

README.md
jaredkhan: No reason no. There is nothing specific to Swift on the JS side so we can do whatever we like. Happy to use Yarn

gulpfile.js
```diff
@@ -0,0 +1,125 @@
...
+
+// Run the Kitura server
+gulp.task('start-server', function (done) {
+  serverProcess = childProcesses.spawn(EXEC_PATH)
+  connectProcessOutput(serverProcess, 'kitura')
+  done()
+})
+
+// All the tasks that handle the building of static files
+gulp.task('static-build', gulp.parallel('styles', 'copy'))
```
moosichu: I get the following error:
```bash
TypeError: gulp.parallel is not a function
    at Object.<anonymous> (/home/tom/Documents/hackersatcambridge/hac-website/ma
```

`gulp --version` gives me:
```bash
[10:36:28] CLI version 3.9.1
[10:36:28] Local version 3.9.1
```

jaredkhan: My mistake, there is an extra step to install Gulp 4 which I forgot about. See updated README

PR #4, opened 2017-03-13
Add Docker support

### LINKED ISSUE #2: Use Yarn instead of NPM
https://yarnpkg.com/lang/en/

Pinpickle:
This PR adds Docker support, so getting up and running with it only requires one thing to install and one command: `docker-compose up`.

I've verified that this actually works, and all of our niceties with live-reloading still work.

Also changes npm to yarn, so it fixes #2


README.md
```diff
@@ -3,18 +3,8 @@
...
-
-Then you will need to install the gulp-cli globally:
-```
-npm rm -g gulp
-npm install -g gulp-cli
-```
-Note: This won't interfere with other Gulp projects you may have on your machine, the new CLI is backwards compatible. For more details on using Gulp 4, see [this blog post](https://www.liquidlight.co.uk/blog/article/how-do-i-update-to-gulp-4/)
-
-Clone this repository
+We use Docker to make this quick and painless for you, [make sure you have it installed](https://docs.docker.com/engine/installation/). Clone this repository
```
moosichu: In my experience, docker can be very fiddly, especially if your version of docker is out if date and you use the distro-provided version. So giving a guaranteed minimum version number is probably ideal.

jaredkhan:
I get the following error which is presumably what is preventing me from seeing some of the previous niceties like highlighted output from the gulp task:

```
debconf: unable to initialize frontend: Dialog
debconf: (Dialog frontend will not work on a dumb terminal, an emacs shell buffer, or without a controlling terminal.)
debconf: falling back to frontend: Readline
debconf: unable to initialize frontend: Readline
debconf: (This frontend requires a controlling tty.)
debconf: falling back to frontend: Teletype
dpkg-preconfigure: unable to re-open stdin: 
```

README.md
Pinpickle: Noted and will leave something in the readme

Pinpickle:
@jaredkhan it's only tangentially related to colours, but I'll fix up colours and that error.

Pinpickle:
@jaredkhan @moosichu I've fixed up the things that you've mentioned.

Just a note: if you've already run `docker-compose up`, you'll need to rebuild with `docker-compose build` to get the new changes.

README.md
```diff
@@ -23,15 +17,25 @@ git clone https://github.com/hackersatcambridge/hac-website.git
...
 ```
 
-followed by:
+This will build all the files and run the web server with appropriate output. It will also watch for changes to the files and rebuild as necessarily. You can stop this process at any time with `^C` (`ctrl` + `C`).
+
+### What `docker-compose up` is doing
+
+This command starts up a docker container that has all the required dependencies, then starts up a development server.
+
+Internally, this command runs the Swift and static file (CSS, images, etc.) build processes in parallel. When they complete, it runs the executable produced by the Swift build. It exposes a server on http://localhost:3000. For convenience, it will reload the browser window for you whenever you make a relevant change to the files. This is powered by BrowserSync, and you can access the interface for it at http://localhost:3001.
```
jaredkhan: Maybe we should mention that it is using Gulp internally. e.g. "Internally, this command uses the Gulp task runner to..." I think the risk with dockerising is that new contributors don't actually know where to start looking when they want to change something because the architecture becomes more opaque. Perhaps worth also pointing people towards the Dockerfile and gulpfile if they want to learn more

Pinpickle: Good one, I'll add a note. I think it's not necessary to point to Dockerfile/gulpfile as that's standard and can be discovered by reading the docs for the relevant technologies.

jaredkhan: Fair point

jaredkhan:
All good my end, this is a mean setup! Once the Linux issues are sorted, 🚀

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN \
+  export DEBIAN_FRONTEND=noninteractive && \
+  apt-get update && \
+  apt-get install curl && \
+  # Allow installation of Node 6: https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions \
+  curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash - && \
+  apt-get update && \
+  apt-get install nodejs && \
+  # Yarn's installation instructions did not work, so we are falling back to npm to install it
+  npm install -g yarn;
```
moosichu: [Yarn advise against this](https://yarnpkg.com/en/docs/install#install-via-npm), strongly:

> Note: Installation via npm is generally not recommended. npm is non-deterministic, packages are not signed, and npm does not perform any integrity checks other than a basic SHA1 hash, which is a security risk when installing system-wide apps.

>For these reasons, it is highly recommended that you install Yarn through the installation method best suited to your operating system.

[Better alternative](https://yarnpkg.com/en/docs/install#linux-tab)

```bash
curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
apt-get update && sudo apt-get install yarn
```

Do we even need nodejs at all with this?

Pinpickle: I'm aware, their instructions (as you provided) didn't work - this is mentioned in the comment.

Node is required as:
- Yarn is written in JS
- Gulp is written in JS

moosichu: Woops, reading file on my part. Sorry!

moosichu:
On Windows, the docker container only seems to be listening on port 8090, but refuses to connect on that port. 

nothing seems to be listening on port 3000

Issue #5, opened 2017-03-14
Don't use npm in order to install yarn


moosichu:
[Yarn advise against this](https://yarnpkg.com/en/docs/install#install-via-npm), strongly:

> Note: Installation via npm is generally not recommended. npm is non-deterministic, packages are not signed, and npm does not perform any integrity checks other than a basic SHA1 hash, which is a security risk when installing system-wide apps.

>For these reasons, it is highly recommended that you install Yarn through the installation method best suited to your operating system.

[Better alternative](https://yarnpkg.com/en/docs/install#linux-tab)

```bash
curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
apt-get update && sudo apt-get install yarn
```

Unfortunately

> Yarn's installation instructions did not work, so we are falling back to npm to install it

moosichu:
Closed by #6

PR #6, opened 2017-03-14
Install Yarn without NPM

moosichu:
Fixes issue #5, desired as it's more secure than npm. (See Issue)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,13 +3,15 @@ FROM ibmcom/swift-ubuntu:3.0.2
 RUN \
   export DEBIAN_FRONTEND=noninteractive && \
   apt-get update && \
-  apt-get install curl && \
+  apt-get install curl apt-transport-https && \
   # Allow installation of Node 6: https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions \
   curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash - && \
   apt-get update && \
   apt-get install nodejs && \
-  # Yarn's installation instructions did not work, so we are falling back to npm to install it
-  npm install -g yarn;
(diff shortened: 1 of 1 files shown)


moosichu:
@Pinpickle Have you tested it? It worked on my machine, but just want to confirm it does on yours!

Dockerfile
```diff
@@ -3,13 +3,15 @@ FROM ibmcom/swift-ubuntu:3.0.2
...
   # Allow installation of Node 6: https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions \
   curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash - && \
   apt-get update && \
   apt-get install nodejs && \
-  # Yarn's installation instructions did not work, so we are falling back to npm to install it
-  npm install -g yarn;
+  # Install Yarn: https://yarnpkg.com/en/docs/install#linux-tab
+  curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+  echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
+  apt-get update && sudo apt-get install yarn
```
Pinpickle: Just had a thought after reviewing, can we do the install of yarn and npm in one apt-get line? This means we'll only have to `apt-get update` twice instead of three times, speeding up the build.

Is `sudo` necessary here?

moosichu: That sudo slipt-me by, and will give it a go!

Pinpickle:
@moosichu I tested it, it is indeed working for me

Dockerfile
Pinpickle: Correction, installation of yarn and _node_

Issue #3, opened 2017-03-12
Setup Travis-CI and unit-testing


moosichu:

moosichu:
Closed by #16

PR #13, opened 2017-03-14
Add Swift testing

jaredkhan:
Addresses issue #3 

Unfortunately for now Swift Package Manager does not allow us to test executable targets which has meant that we require some restructuring for this to work. This is not ideal, I expect this will not be the case in 6 months time but it's not a huge hassle to change this later on. 

I've set up an example Swift test case. You can run this with `swift test` or `yarn test`  in the container. Tests currently pass.


Sources/HaCWebsiteLib/routes.swift
```diff
@@ -0,0 +1,72 @@
...
+    let description: String
+    let prerequisites: String
+  }
+
+  router.get("/events") { request, response, next in
+    defer {
+      next()
+    }
+    do {
+      var context: [String: Any] = [
```
moosichu:
```
/root/hac-website/Sources/HaCWebsiteLib/routes.swift:56:11: warning: variable 'context' was never mutated; consider changing to 'let' constant
      var context: [String: Any] = [
      ~~~ ^
```

From Travis, might be worth changing.

jaredkhan:
@moosichu I've fixed the warning that was there. I'd like to disallow warnings but the warnings that the dependencies are throwing might make it fiddly, I'll make a separate issue for this and we can also discuss the implications of disallowing warnings there.

PR #22, opened 2017-03-31
Add workshop parsing, migrate to Swift 3.1

jaredkhan:
This PR has a lot of changes in it so I apologise now for the lack of granularity.

Includes:
- Migration to Swift 3.1
- Parsing of workshop files
- Testing of workshop files
- Setup for test resources (e.g. example workshop descriptions with known expected parse outputs)
- Fixes to the gulpfile
- A .env setup

Some points for discussion:
- Is the .env setup presented here robust enough? Currently only check for value existence on startup
- Is the direct `git pull` approach here the best way for keeping workshops up-to-date? I'd argue it's highly usable as it avoids us forgetting to manually update, especially useful when other people start pushing to the workshops repo. It does result in some complications when changing the schema though


Tests/HaCWebsiteLibTests/HaCWebsiteLibTests.swift
```diff
@@ -2,16 +2,94 @@ import XCTest
 @testable import HaCWebsiteLib
 
 class HaCWebsiteLibTests: XCTestCase {
-    func testExample() {
```
Pinpickle: Move out of the way, we have _real tests_ coming on in 🙌

docker-compose.yml
```diff
@@ -13,7 +13,6 @@ services:
     - "3000:3000"
     - "3001:3001"
     volumes:
-    - Packages:/root/hac-website/Packages
```
Pinpickle: If we're not using this volume anymore, can we get rid of the definition of the volume in this file as well?

Tests/HaCWebsiteLibTests/HaCWebsiteLibTests.swift
```diff
@@ -2,16 +2,94 @@ import XCTest
...
+      throw TestError.couldNotLoadFile("path")
     }
+  }
+
+  func testWorkshopParse() throws {
+    let testMetadata = try getTestResourceString(from: "conquering_linux/metadata.yaml")
+    let testDescription = try getTestResourceString(from: "conquering_linux/description.md")
+    let testPrerequisites = try getTestResourceString(from: "conquering_linux/prerequisites.md")
+
+    let workshop = try parseWorkshop(metadataYaml: testMetadata, descriptionMarkdown: testDescription, prerequisitesMarkdown: testPrerequisites)
```
Pinpickle: It looks like this function doesn't exist?

Tests/HaCWebsiteLibTests/HaCWebsiteLibTests.swift
```diff
@@ -2,16 +2,94 @@ import XCTest
...
+
+  /**
+   * Gets the absolute path to a resource in HaCWebsiteLibTests/Resources directory.
+   * Generates an XCTest failure and throws an error if the resource cannot be found
+   *
+   * - returns:
+   * A String of the absolute path to the resource
+   *
+   * - throws:
+   * An error of type TestError.fileNotFound if the fil cannot be found
```
Pinpickle: Nit: Spelling mistake (`fil/file`)

Tests/HaCWebsiteLibTests/HaCWebsiteLibTests.swift
```diff
@@ -2,16 +2,94 @@ import XCTest
...
+    let path = try getTestResourcePath(at: relativePath, file: file, line: line)
+    do {
+      return try String(contentsOfFile: path, encoding: .utf8)
+    } catch {
+      self.recordFailure(withDescription: "Could not load file at \(path)", inFile: file, atLine: line, expected: false)
+      throw TestError.couldNotLoadFile("path")
     }
+  }
+
+  func testWorkshopParse() throws {
```
Pinpickle: It looks like this test is testing for a lot of different things. Would it be advantageous to split it into a few smaller tests, so that when they fail we get a better understanding of what has gone wrong?

Tests/HaCWebsiteLibTests/HaCWebsiteLibTests.swift
```diff
@@ -2,16 +2,94 @@ import XCTest
...
+   *
+   * - throws:
+   * An error of type TestError.fileNotFound if the fil cannot be found
+   *
+   * - parameters:
+   *   - at: The path of the desired resource relative to HaCWebsiteLibTests/Resources
+   *   - file: Optional file name for better error reporting
+   *   - line: Optional line number for better error reporting
+   */
+  private func getTestResourcePath(at relativePath: String, file: String = #file, line: UInt = #line) throws -> String {
```
Pinpickle: Moving this test setup code into reusable functions is +++

Tests/HaCWebsiteLibTests/HaCWebsiteLibTests.swift
```diff
@@ -2,16 +2,94 @@ import XCTest
 @testable import HaCWebsiteLib
 
 class HaCWebsiteLibTests: XCTestCase {
```
Pinpickle: It seems to me like this file isn't testing `HacWebsiteLib` but just `Workshop`. Should the name of the file reflect that?

Sources/HaCWebsiteLib/routes.swift
```diff
@@ -33,38 +28,30 @@ public func serveWebsite() {
       next()
     }
     do {
-      try response.render("workshops", context: [:]).end()
+      let workshops = WorkshopManager.workshops
+      dump(workshops)
```
Pinpickle: Are these dumps supposed to be here?

Tests/HaCWebsiteLibTests/HaCWebsiteLibTests.swift
Pinpickle: In particular, this test is testing that file reading works, yaml parsing works and that our mapping from the yaml data structure to our internal data structure works.

Sources/HaCWebsiteLib/Workshop.swift
```diff
@@ -0,0 +1,103 @@
...
+// ...and for Workshop.Thanks
+func ==(lhs: Workshop.Thanks, rhs: Workshop.Thanks) -> Bool {
+  return lhs.to == rhs.to && lhs.reason == rhs.reason
+}
+
+public extension Workshop {
+  /**
+    Parses the files that describe a workshop into a Workshop instance
+
+    - returns:
```
Pinpickle: All of these doc blocks are tasty

Sources/HaCWebsiteLib/Workshop.swift
```diff
@@ -0,0 +1,103 @@
...
+    let title: String
+    let url: URL
+  }
+
+  enum WorkshopError : Error {
+    case malformedMetadata
+  }
+}
+
+// Satisfy equatability for Workshop.Recommendation
```
Pinpickle: Are these comments necessary?

Sources/HaCWebsiteLib/GitUtils.swift
```diff
@@ -0,0 +1,76 @@
+import Foundation
+
+public struct GitUtils {
+  /**
+    Executes a shell command synchronously, returning the exit code
+    
+    - parameters:
+      - in: The path to the directory in which to execute the command
```
Pinpickle: It looks like this `in` parameter doesn't exist?

Sources/HaCWebsiteLib/GitUtils.swift
```diff
@@ -0,0 +1,76 @@
...
+    // Note the Objective-C gunk here, an unfortunate part of this API
+    var isDirectory: ObjCBool = ObjCBool(false)
+    let fileExists = FileManager.default.fileExists(atPath: localRepoPath, isDirectory: &isDirectory)
+
+    if fileExists && isDirectory.description == "true" {
+      // Directory already exists, pull updates
+      pull(in: localRepoPath)
+    } else {
+      // Directory does not yet exist, clone the repo
+      clone(repoURL: "https://github.com/hackersatcambridge/workshops.git", in: localPath, directory: directory)
```
Pinpickle: Is it possible to do a shallow/single branch clone to prevent downloading more than we need?

https://www.perforce.com/blog/141218/git-beyond-basics-using-shallow-clones
http://stackoverflow.com/questions/1778088/how-to-clone-a-single-branch-in-git

Sources/HaCWebsiteLib/WorkshopManager.swift
```diff
@@ -0,0 +1,68 @@
...
+  public private(set) static var workshops: [Workshop] = []
+
+  /// Used to avoid multiple git pulls being run simultaneously
+  private static let serialQueue = DispatchQueue(label: "com.hac.website.ws-pull-queue")
+
+  /**
+    Ensures that an up-to-date copy of the workshops repo is available locally
+    We don't want this function running many times concurrently, hence 'unsafe'
+  */
+  private static func unsafePull() {
```
Pinpickle: To reflect the corresponding `update` method, might `unsafeUpdate` be a better name?

Sources/HaCWebsiteLib/WorkshopManager.swift
```diff
@@ -0,0 +1,68 @@
+import Foundation
+import Dispatch
+import DotEnv
+
+public struct WorkshopManager {
+  /// The list of workshops as derived from the workshops directory last time it was updated
```
Pinpickle: Wasn't aware that this triple slash syntax was a thing

#learning

README.md
```diff
@@ -16,8 +16,13 @@ Instructions are as follows. Clone this repository:
 ```bash
 git clone https://github.com/hackersatcambridge/hac-website.git
 ```
+Navigate to the hac-website directory that you've now cloned. Create a file named `.env` and give it the following contents:
+```
+DATA_DIR=/root/hac-website/Data
+```
+Save the file. Why did we have to do that? Leaving this file out of our source control means that we can store sensitive information like API keys in here for our server to use.
```
Pinpickle: In this case it isn't sensitive information, but environment specific information (DATA_DIR will be different on Heroku)

README.md
```diff
@@ -16,8 +16,13 @@ Instructions are as follows. Clone this repository:
 ```bash
 git clone https://github.com/hackersatcambridge/hac-website.git
 ```
+Navigate to the hac-website directory that you've now cloned. Create a file named `.env` and give it the following contents:
+```
+DATA_DIR=/root/hac-website/Data
```
Pinpickle: Would it be better to set this config through our `docker-compose` file? This means less setup and people can continue to use the `.env` setup on their Macs because `DATA_DIR` will need to be different for container vs macOS.

Sources/HaCWebsite/main.swift
```diff
@@ -1,3 +1,15 @@
 import HaCWebsiteLib
+import DotEnv
+import Foundation
 
-serveWebsite()
+DotEnv.load()
+
+// Check for required env vars
+guard let _ = DotEnv.get("DATA_DIR") else {
+  fatalError("Missing environment variable")
```
Pinpickle: Should we colocate these checks with the code that uses the variable? e.g. in `WorkshopManager`. We could add a `requireEnvVar` to our [`Config`](https://github.com/hackersatcambridge/hac-website/blob/master/Sources/HaCWebsiteLib/configuration.swift) struct as well.

Sources/HaCWebsiteLib/WorkshopManager.swift
Pinpickle: I see, the `update` also processes the workshop files. Carry on

Sources/HaCWebsiteLib/WorkshopManager.swift
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  /**
+    Update the workshops array by pulling changes from the repo.
+    This should be called whenever it is likely that changes
+    have been made to the repo.
+  */
+  public static func update() throws {
+    // Use a serial queue to avoid concurrent Git processes
+    try serialQueue.sync {
```
Pinpickle: Is this necessary? I don't fully understand concurrency in Swift but I thought that we didn't have concurrency yet?

Which raises another question: will this block requests while the repo is being updated?

Package.pins
```diff
@@ -0,0 +1,120 @@
+{
+  "autoPin": true,
```
Pinpickle: :tada:

Sources/HaCWebsiteLib/WorkshopManager.swift
jaredkhan: We do indeed have concurrency. Dispatch is the chief concurrency API in Swift and it has been ported to Linux though progress remains to be made on integrating it into the Linux port of the Foundation library. You can dig through more here: https://github.com/apple/swift-corelibs-libdispatch/

 Kitura uses Dispatch for request handling. When a new request comes in, a new task (the code we write in the route handler) is added to a concurrent DispatchQueue. This means that we are safe to have multiple visitors to our site at once and that this code here does not block requests while the pull is in progress, only the request to /api/refresh_workshops

Sources/HaCWebsiteLib/Workshop.swift
jaredkhan: I'd argue it's not clear that they are here for satisfying the Equatable protocol otherwise.

Sources/HaCWebsiteLib/GitUtils.swift
jaredkhan: Good catch, changed this later on. Fun story, the currentDirectoryPath property on Process does absolutely nothing on Linux and that's not documented anywhere 🤷‍♀️

README.md
jaredkhan: 👍 Will move this to the `docker-compose` but since that is loaded before the `.env`, the `.env` will overwrite the docker-compose so we can have to have a think later about how to support multiple environments here

Tests/HaCWebsiteLibTests/HaCWebsiteLibTests.swift
jaredkhan: 👍 I was a little shaky on the structure of these test cases but have now structured it more logically

Tests/HaCWebsiteLibTests/HaCWebsiteLibTests.swift
jaredkhan: It does look like a lot but really it's only testing if the mapping to our internal structure works. The rest is handled by libraries. Certainly once we have a more concrete Yaml format it will be good write some more fine-grained tests for this.

jaredkhan:
Okay @Pinpickle, give this a whirl now

Sources/HaCWebsite/main.swift
jaredkhan: That would be nice but we want to fail fast (I.e on load) if the environment variable doesn't exist. Config struct sounds like a good option

Pinpickle: We can still fail hard and fast even when the check is colocated with the code that uses it - just needs to be done statically. 

Alternatively, we could make it optional (to make getting started easier) and make a note on `/workshops` when the repository is not loaded. This isn't necessarily the best example, but every developer is not going to need every feature of our website.

Sources/HaCWebsiteLib/Workshop.swift
Pinpickle: It feels like we're locking ourselves into writing a comment explaining this every time we need to conform to that protocol. This opens us up to people forgetting, convention on the comments, multiple sources of truth etc...

This seems like pretty idiomatic Swift to me, and as a non-Swiftie, was fairly clear.

However, if clarity is a concern, is it clearer to use the example that Apple gives on the docs? https://developer.apple.com/reference/swift/equatable shows that we can add the comparison as a static function on the struct:

> To adopt the Equatable protocol, implement the equal-to operator (==) as a static method of your type.

This will bring the equality function much closer to the struct and make the intent clearer.

Sources/HaCWebsiteLib/Workshop.swift
```diff
@@ -0,0 +1,103 @@
...
+      - prerequisitesMarkdown: The String contents of the Markdown file with the prerequisites of this workshop
+   
+    - throws:
+      An error of type Workshop.WorkshopError.malformedMetadata if the Yaml file does not follow the expected structure
+   
+   */
+  public static func parse(metadataYaml: String, descriptionMarkdown: String, prerequisitesMarkdown: String) throws -> Workshop {
+    // Parse Yaml and Markdown
+    let metadata = try Yaml.load(metadataYaml)
+    let description = KituraMarkdown.render(from: descriptionMarkdown)
```
Pinpickle: Do we want to store the markdown rendered? Having it as markdown or as some other format is pretty useful as it is not uncommon to need plaintext as opposed to domain-specific html.

jaredkhan: Good point. We could store both, the idea of avoiding re-rendering the markdown for every serve is appealing to me, do you think it's too domain-specific to store at all?

Pinpickle: Pre-rendering for performance seems like premature optimisation

Sources/HaCWebsiteLib/WorkshopManager.swift
Pinpickle: Awesome, thanks for clarifying!

Sources/HaCWebsiteLib/routes.swift
```diff
@@ -33,38 +28,29 @@ public func serveWebsite() {
       next()
     }
     do {
-      try response.render("workshops", context: [:]).end()
+      let workshops = WorkshopManager.workshops
+      dump(workshops.map {($0.description.html, $0.prerequisites.html)})
```
Pinpickle: Do we still want to dump this info on every request?

jaredkhan: Foiled again, can we make Travis shout at me for this?

Tests/HaCWebsiteLibTests/HaCWebsiteLibTestCase.swift
```diff
@@ -0,0 +1,63 @@
+import Foundation
+import XCTest
+
+class HaCWebsiteLibTestCase: XCTestCase {
```
Pinpickle: I like that we're creating reusable test code, but is there any reason why we're using inheritance here? (e.g. is there protected state that we need to access for these to work) In general, inheritance adds incidental complexity and from what I can tell, is not very Swift-ey outside of the context of iOS.

I know we need to inherit from `XCTestCase` to make tests work, but we don't need to add any more complexity if we don't need to.

Tests/HaCWebsiteLibTests/WorkshopTests.swift
```diff
@@ -0,0 +1,81 @@
...
+
+  func testMissingTitle() {
+    checkThrows(metadataPath: "missing_title.yaml")
+  }
+
+  func testMissingThanksTo() {
+    checkThrows(metadataPath: "missing_to.yaml")
+  }
+
+  // Should catch
```
Pinpickle: Don't the names of the tests enumerate exactly what the following comments explain?

Tests/HaCWebsiteLibTests/HaCWebsiteLibTestCase.swift
jaredkhan: I agree it's non-ideal but I'm struggling to find a tidier option. The reason in this case is for the use of self.recordFailure, a method on XCTestCase which is the thing that actually triggers the failure. Granted, the test would fail if we just had throwing functions, but it's not as clear in the report what actually went wrong. Alternatives welcomed

Tests/HaCWebsiteLibTests/WorkshopTests.swift
```diff
@@ -0,0 +1,81 @@
+import XCTest
+@testable import HaCWebsiteLib
+
+class WorkshopTests: HaCWebsiteLibTestCase {
```
Pinpickle: This test suite is looking really nice!

PR #29, opened 2017-07-19
BEMify existing styles

Pinpickle:
This moves existing styles to a BEM style for class names and variable names. Variable names using BEM isn't really an official thing but my own invention I've adopted in other code bases. As CSS preprocessors use variables in the global scope, it helps us keep that scope manageable.


static/src/styles/landing/LandingUpdateFeed.styl
```diff
@@ -0,0 +1,7 @@
+.LandingUpdateFeed {
```
jaredkhan: Does each Component need its own file?

Views/home.stencil
```diff
@@ -3,27 +3,27 @@
 {% block title %}Hackers at Cambridge{% endblock %}
 
 {% block content %}
-  <header class="">
+  <header>
     <h1>Hackers at Cambridge</h1>
-    <h2 class="mono">Cambridge's student tech society</h2>
```
jaredkhan: Is there a way to handle universal things like 'mono' with BEM?

static/src/styles/core/color.styl
```diff
@@ -0,0 +1,7 @@
+$Color__light = #F2F2F2;
```
jaredkhan: Any rules we should know around these prefixes?

static/src/styles/main.styl
```diff
@@ -1,41 +1,2 @@
...
-.post_type {
-  display: inline-block;
-  background: $lightBackground;
-  padding: 0.3em;
-  color: $bright1;
-  border-radius: $smallBorderRadius;
-}
\ No newline at end of file
+@require './core';
+@require './landing';
```
jaredkhan: Is the approach to have all pages read from the same single css file?

Views/home.stencil
Pinpickle: We can handle this in a few ways (which are not mutually exclusive). The key thing to remember here is that `mono` does not semantically describe what the class does, it describes the style that it applies, which tightly couples the semantics to the styling. 

We're giving this text a monospaced font because of the uniform space taken up by each glyph, but to apply some kind of emphasis.

This is especially important if the person writing the markup is also not the designer (mono means something significant to designers but maybe not to others).

So the key is to make a semantic class name for this. There may be more than one of these.

For example:

- We make a general `Text--emphasis` class (this probably isn't the best name due to confusion with `<em>`, but you get the idea, that does exactly what `mono` did.
- Give this particular `h2` element a very descriptive semantic name, `.LandingIntro__tagline`, and apply the styles directly to that. We could make a mixin (maybe called `Text--emphasis`?)

And there are options in between as well.

Again, this is similar to the process you'd go with deciding to name functions/variables/classes in the rest of your code. Do we name a function based on the algorithm it uses, or what you'd actually use the function for?

static/src/styles/landing/LandingUpdateFeed.styl
Pinpickle: No requirement here, I'm just trying to split things up where possible.

static/src/styles/core/color.styl
Pinpickle: Nothing hard and fast, just naming the prefix after the file it's contained in, maybe this file should be called `Color` instead of `color` for consistency with other components.

As I said, BEM doesn't really have anything to say about these variables. I've conjured this idea that `Color` is a block, and `light`, `bright`, etc are all elements of it.

static/src/styles/main.styl
Pinpickle: Not a BEM thing, just for our own sanity.

Do you think requiring a set of CSS files would on a per-page basis would be better?

Views/home.stencil
jaredkhan: Makes a lot of sense, thanks for explaining this!

static/src/styles/main.styl
jaredkhan: Not necessarily, this project will remain small enough for this to be a pretty sensible option, just checking I understood 😉

static/src/styles/landing/LandingUpdateFeed.styl
jaredkhan: Nice! We'll see if this slows us down and can make changes if need be.

jaredkhan:
Looks great! This pattern is so intuitive. 🚀

PR #30, opened 2017-07-19
Flesh out core styles

Pinpickle:
Depends on #29

This fleshes out some of the empty placeholders left in the BEM PR, and adds some well-needed defaults (normalize.css, sensible typography)

Is still by no means pretty, we'll get there.

image


gulpfile.js
```diff
@@ -21,7 +21,7 @@ gulp.task('copy', () =>
 
 // Process the Stylus files into CSS
 gulp.task('styles', () =>
-  gulp.src('static/src/styles/**/*.styl')
+  gulp.src('static/src/styles/main.styl')
```
jaredkhan: Won't this only pick up changes to main and not recompile changes in other files?

static/src/styles/core/fonts.styl
```diff
@@ -0,0 +1,3 @@
+// This file must be imported before any files that contain CSS output
+// This is because @import declarations must be the first declarations in CSS files
```
jaredkhan: With this in mind, is there a better way we should be importing fonts or is this fine?

static/src/styles/core/layout.styl
```diff
@@ -0,0 +1,27 @@
...
+
+body {
+  background: $Color__lightBackground;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+// We are violateing BEM naming rules here purely because this function is going to be used a lot
```
jaredkhan: Bro do you even spell
`violateing`

Pinpickle: It's 🥝 spelling I swear

gulpfile.js
Pinpickle: Nah, this doesn't affect the watch code (which is further down), a recompile is still triggered if any file is changed.

static/src/styles/core/fonts.styl
Pinpickle: Putting the `link` directly in our HTML will prevent us from accidentally breaking the fonts

gulpfile.js
jaredkhan: Can see that this works but I don't understand why. My understanding is that `gulp.src` will output all files which match the given glob which in this case is just this one file.
I thought `changed(dest)` just filtered out its input to those files that have changed, how is it finding the changed files?

jaredkhan:
Jeeeez. Dunno what happened to this merge.

jaredkhan:
@Pinpickle Did you rebase a commit that was already merged 👹

gulpfile.js
Pinpickle: I believe the `changed` isn't doing anything in this case as by default it compares hashes of files, not modification dates. `main.styl` will always be different to `main.css`. I'm also not sure if it knows to change file extensions like that by default, either.

jaredkhan:
This has been merged into master, apologies for the confusion

PR #31, opened 2017-08-21
Implement header design

jaredkhan:
Implements the design that we've been carving out for the header of the landing page:

screen shot 2017-08-20 at 20 42 44


static/src/styles/landing/LandingIntroduction.styl
```diff
@@ -1,3 +1,30 @@
...
+  color: $Color__mediumText;
+}
+
+.LandingIntroduction__link {
+  text-decoration: none;
+  color: $Color__brightText;
+}
+
+.LandingIntroduction__link:after {
+    content:' ➝';
```
Pinpickle: Now this is some semantic beauty right here 😍

Views/home.stencil
```diff
@@ -3,15 +3,24 @@
...
-  </header>
-  <section class="LandingFeature">
-    <h2>Featured</h2>
-    <p>Hack Cambridge</p>
-  </section>
+  <div class="LandingTop">
+    <header class="LandingIntroduction LandingTop__Introduction">
+      <div class="LandingIntroduction__titleContainer">
+        <img src="static/images/hac-logo-dark.svg" alt="Hackers at Cambridge" class="LandingIntroduction__titleImage">
+        <h4 class="LandingIntroduction__tagLine">Cambridge's student tech society</h4>
```
Pinpickle: `hx` tags are supposed to be headings of sections, but here (and I've seen this elsewhere), it's being used stylistically (`h4` is about the right size that you want).

We might want to make this the correct tag semantically (e.g. in a `p` element instead, but a `div` usually works just fine). Elsewhere you've used the `Text__headerScale` mixin which I think is a good way to achieve your desired convenience while also keeping things semantic.

You can also make a `Text--h4` modifier if you're really after some quick-to-hand convenience.

static/src/styles/core/color.styl
```diff
@@ -1,12 +1,15 @@
 $Color__light = #F2F2F2;
+$Color__medium = #878787;
```
Pinpickle: What exactly is this a `medium` of? 🤔

Pinpickle:
This is awesome! I left a couple of pretty small comments but otherwise this is pretty true to BEM methodology.

static/src/styles/core/color.styl
jaredkhan: Any ideas for this? 'mediumGrey'? 'mediumBrightnessMuted'? 'fadedTextForLightBG'

Pinpickle: There isn't a great answer here, and we don't need to break our brains figuring it out so `mediumGray` sounds good to me (note the US spelling :cry:)

Pinpickle:
:shipit:

PR #32, opened 2017-08-22
Style the updates feed

jaredkhan:
Executes a fairly literal interpretation of our updates feed design. This one got a bit hairy. Trying to find a tidy way of aligning the action at the bottom of the cards didn't go well nor did trying to find a way to make the grid not have outside padding.

screen shot 2017-08-22 at 01 07 15

PR #46, opened 2017-09-02
Landing Page Controller + Update Feed styling

jaredkhan:
Sorry for tangling these together. Will take care to not do this in future.

Here, we define a controller that manages the data for the landing page and sends it to view models. 
We also style the updates feed based on discussion in #32 

I think this is a pretty nice set up. I like the idea of having view models that match up with BEM blocks though I haven't gone fully that way yet. 

I really like the way we can separate out our components here. It could be useful to make some sandbox route in the dev environment e.g. `/sandbox/PostCard` that allows us to see components in isolation


Sources/HaCTML/Attributes.swift
```diff
@@ -1,13 +1,22 @@
 private func stringAttribute(_ key: String) -> AttributeKey<String> {
-  return .init(key, apply: { .text($0) })
+  return AttributeKey(key, apply: { .text($0) })
```
jaredkhan: I kept stumbling on trying to read this `.init`

Sources/HaCTML/Attributes.swift
```diff
@@ -1,13 +1,22 @@
...
 }
 
 private func numberAttribute(_ key: String) -> AttributeKey<Int> {
-  return .init(key, apply: { .text(String($0)) })
+  return AttributeKey(key, apply: { .text(String($0)) })
 }
 
-private func CSSAttribute(_ key: String) -> AttributeKey<[String: String]> {
-  return .init(key, apply: { .text($0.map({ "\($0): \($1);" }).joined(separator: " ")) })
+private func CSSAttribute(_ key: String) -> AttributeKey<[String: String?]> {
```
jaredkhan: 50/50 on whether this is a good idea. Fundamentally this attribute deals with Strings to Strings but having this value be optional made things easier at the point of use (for me)

Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
```diff
@@ -0,0 +1,53 @@
...
+  ]
+
+  static var handler: RouterHandler = { request, response, next in
+    defer {
+      next()
+    }
+    do {
+      try response.send(
+        LandingPage(updates: updates).node.render()
+      ).end()
```
jaredkhan: `end` can throw a Socket.Error
We should decide on a way to stop every controller from having to catch this

Sources/HaCTML/Attributes.swift
Pinpickle: Good call, should we give the other attribute functions the same treatment?

Sources/HaCTML/Attributes.swift
Pinpickle: I share your sentiment. This should be a concern if an optional value can sneak its way in there unintentionally. Maybe we can come up with a friendly way to say "yeah this is optional but I'm doing it on purpose", so the behaviour is explicit?

Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
Pinpickle: How about extending the response type to have a `safelyEnd` method that catches and logs this error? While we're at it we could overload `send` to allow a `Node` type (too magical?)

Sources/HaCTML/Attributes.swift
Pinpickle: I spoke to soon, you did exactly that!

static/src/styles/landing/LandingUpdateFeed.styl
```diff
@@ -1,14 +1,35 @@
...
+  padding-left: grid(4);
 }
 
 .LandingUpdateFeed__item {
-  padding: grid(2);
-  flex: 1 1 auto;
-  width: grid(40);
+  padding: grid(1);
+  display: flex;
+  width: 25%;
```
Pinpickle: Check out those gutters!

Sources/HaCWebsiteLib/ViewModels/PostCard.swift
```diff
@@ -0,0 +1,32 @@
...
+    case hackathon = "Hackathon"
+  }
+
+  let title: String
+  let category: PostCategory
+  let description: String
+  let backgroundColor: String?
+  let imageURL: String?
+  
+  var node: Node {
```
Pinpickle: This is awesome!

jaredkhan: Once we have made a few more components and have played with different patterns we can think about things like `ViewModel` protocols etc if we want to

PR #52, opened 2017-09-07
Fix node engine version to accept above 6.10.0

CharlieCrisp:
This PR fixes an issue which was preventing people from cloning the repo and starting up the server properly unless they had a docker image with a specific node version installed.


CharlieCrisp:
This was closed and recreated because the work was moved to a separate branch

PR #53, opened 2017-09-07
Fix node engine version to use 6.11.3

CharlieCrisp:
This PR fixes an issue which was preventing people from cloning the repo and starting up the server properly unless they had a docker image with a specific node version installed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "heroku-postbuild": "gulp static-build"
   },
   "engines": {
-    "node": "6.10.0"
+    "node": "6.11.3"
   },
   "author": "",
   "license": "ISC",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "heroku-postbuild": "gulp static-build"
   },
   "engines": {
-    "node": "6.10.0"
+    "node": ">=6.10.0"
```
jaredkhan: This is going to allow any later version of Node to work with this. Is this what we want? We can control the version of Node that gets installed using the Dockerfile

Pinpickle: I think that pinning it to a specific version would be a good idea as well, it will reduce "works on my machine" scenarios.

It's worth noting that this `>=` will also push us up to `8.x` which is crossing major version boundaries which will introduce breaking changes.

Finally, `6.x` versions prior to `6.11.1` have a security vulnerability (https://nodejs.org/en/blog/vulnerability/july-2017-security-releases/), and while it's unlikely to affect us (our Node instance is not exposed to the big bad internet), we have an opportunity to fix it here as well.

CharlieCrisp: So this change is actually oudated now and I've taken a different approach. It pins it to a specific version but as you mentioned it will currently be affected by the security vulnerability.

Pinpickle: My bad, didn't notice that this was outdated!

PR #55, opened 2017-09-14
Add database

jaredkhan:
Adds a PostgreSQL database to our Docker setup.

**Remaining to be done:**
- Set up a Postgres instance on Heroku
- Better management of the creds so that we can actually use it


Sources/HaCWebsiteLib/routes.swift
```diff
@@ -44,3 +47,28 @@ public func serveWebsite() {
...
+  do {
+    let driver = try Driver(
+      masterHostname: "hac-db",
+      readReplicaHostnames: [],
+      user: "richard",
+      password: "test",
+      database: "hac"
+    )
+    Database.default = Database(driver)
+    try Pet.prepare(Database.default!)
```
Pinpickle: It appears here that we're setting a default drive _and_ setting it explicitly (part of the reason why we have a `!` here)

Because we're setting a default, can we omit it here?

docker-compose.yml
```diff
@@ -9,12 +9,22 @@ services:
...
+      - node_modules:/root/hac-website/node_modules
+      - yarn_cache:/yarn_cache
+      - .:/root/hac-website/
+    depends_on:
+      - hac-db
+  hac-db:
+    image: postgres
+    environment:
+      POSTGRES_PASSWORD: test
+      POSTGRES_USER: richard
```
Pinpickle: Bless his soul

Pinpickle:
@jaredkhan I'm not considering any comments on code blockers, but can we figure out what's causing the test failures?

jaredkhan:
Looks like something is already tying up port 5432, will take a closer look later on

jaredkhan:
Turns out the trusty travis image already has a Postgres instance running by default so the port is tied up. Could have killed that instance and waited for the port to be released but given that we're only exposing this port for dev convenience, think we should just use this non standard port 5433

Dockerfile
```diff
@@ -13,7 +13,9 @@ RUN \
   apt-get update && \
   apt-get install nodejs yarn && \
   npm install -g n && \
-  n 6.11.3
+  n 6.11.3 && \
+  # Install PostgreSQL library
```
moosichu: Naughty

Issue #35, opened 2017-08-28
Define and build normalised stream of "updates"


Pinpickle:
Our home page features an update stream. It has upcoming events as well as things we've done (videos, blog posts, all the good stuff).

We need a way to combine these two concepts in a way that doesn't result in one being dwarfed by the other (events are in the future, released things are in the past).

Events can be pulled in by #23 but other updates need some kind of representation in our database.

CharlieCrisp:
My thoughts on the matter:
Stuff that we have done in the past will be fairly easy to filter on a newest to oldest basis.

In terms of events, we need a way of filtering which events are relevant at any given time. Provided this filtering works well, we can prioritise the relevant events without swamping the past projects. This assumes that we will never have more than roughly 2-3 relevant events at once, which I think is a fair assumption for about 90% of the time.

Unfortunately, filtering is not so easy because some events will be important even when they are a few months away (e.g. ChariTech) whereas workshops are only relevant in the week before.

I suggest that we perhaps define a notion of a warning period and TTL for events. I.e. two figures we input for each event which describe how far before and after the event they should be advertised.
We can then calculate if we are in the 'live period' for each workshop and use this info to choose whether we display it or not.

Just a suggestion but would be interested to hear thoughts about this :)

Pinpickle:
@CharlieCrisp that looks like it will give us a good amount of freedom going forwards

While this solves the when to display issue, it doesn't solve the sorting issue. How do we intersperse future things with past things?

CharlieCrisp:
@Pinpickle So my thinking was that we show the future things first and then fill the remainder with past things. If we showed 4 things in total then this may mean we get swamped with future events but only when we're very busy. Alternatively we could show like 8 events and then we'd definitely have a good mix.

moosichu:
Will we have dedicated page to view an unlimited number of these "updates" filtered in an arbitrary fashion?

Pinpickle:
@moosichu that is probably useful, but I don't think is necessary for MVP.

moosichu:
Can this issue be closed now?

Pinpickle:
@moosichu I don't believe so - we have figured out events but not really any of the other things.

moosichu:
This issue depends on a load of others, can someone please try and find them all and list them here?

jaredkhan:
Tasks within this include: 
- Designing the data structure of an 'Update'
- Creating a controller that takes many stream providers and outputs a stream of updates
[...]

Issue #23, opened 2017-04-03
Need a database representation of 'Events'


jaredkhan:
'Events' are things that are advertised with a time and date and any other additional information necessary. We need a representation of events for the website that allows for referencing of `Workshop`s and Facebook events.

Pinpickle:
Ideas for things that should be in events:

- Cover image
- Link to FB event
- Link to workshop page (or with contents embedded)
- Location
  - As "Google Maps aware" with address and location
  - As "building aware" with room number etc
- Time
- Description that is bound to the event and not the workshop (e.g. "there will be pizza", "happy halloween", etc...)

CharlieCrisp:
@Pinpickle currently we've also got a `category` section for our events' `Postcard`s. I think this is probably also a good idea for `Event`s to have. 

Also, why the distinction between general event description and technical description? Is there a benefit of keeping these separate?

Pinpickle:
@CharlieCrisp maybe the `category` for events needs to be a subset of those available for `PostCard`? While `hackathon` and `workshop` are useful for events, I can't think of a case for `hackathon` (and thinking ahead, `blog-post`)

Are you asking for why we're separating workshops from events? The goal is to make workshops completely self-standing as it is possible for us to run the same workshop more than once. This will allow us to reuse that content, and also make the workshop pages useful in their own right all year round.

Another way of putting it:

- The audience for our event pages is people in Cambridge
- The audience for our workshop pages is anyone in the world interested in CS

CharlieCrisp:
One thing we haven't considered here is the database representation of past projects e.g. videos, workshop videos, blogs etc. 

For these things, we would not need `TTL`s etc. but we may need other information?

jaredkhan:
Videos and blogs should probably be independent of this, since they aren't events. Perhaps they can have publication dates tied to them and they become 'updates' and can be added to an update stream.

PR #56, opened 2017-09-17
Introduce representation of Events

CharlieCrisp:
This PR is a WIP that introduces the notion of Events. 

It removes the hardcoded `Event` `PostCard`s on the landing page and replaces them with ones which are sourced from and `EventManager`

This relates to #23 and #35 

Feedback Please!! 😄

### DIFF
--- a/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
+++ b/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
@@ -7,36 +7,22 @@ import DotEnv
 import SwiftyJSON
 
 struct LandingPageController {
-  static let updates = [
-    PostCard(
-      title: "Conquering Linux",
-      category: .workshop,
-      description: "Be great at computers, not because you can, but because you dream.",
-      backgroundColor: "purple",
-      imageURL: "/static/images/workshop.jpg"
-    ),
+
+  static let eventPostCards : [PostCard] = EventManager.events.flatMap { event in
+    return event.toPostCard()
+  }
+
+  static let videos = [
     PostCard(
       title: "Partial Recursive Functions 1: Functions",
       category: .video,
       description: "Learn all the things.",
       backgroundColor: "#852503",
       imageURL: "/static/images/functions_frame.png"
-    ),
-    PostCard(
-      title: "Intermediate Git",
-      category: .workshop,
-      description: "Merge conflicts? Rescuing yourself from a hairy rebase? No sweat.",
-      backgroundColor: "black",
-      imageURL: "/static/images/intermediate_git.png"
-    ),
-    PostCard(
-      title: "Game Gig 2017",
-      category: .hackathon,
-      description: "Are you game?",
-      backgroundColor: "#ECA414",
-      imageURL: nil
     )
   ]
+
+  static let updates = eventPostCards + videos
 
   static var handler: RouterHandler = { request, response, next in
     defer {
--- a/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
+++ b/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
@@ -19,6 +19,13 @@ struct LandingPageController {
       description: "Learn all the things.",
       backgroundColor: "#852503",
       imageURL: "/static/images/functions_frame.png"
+    ),
+    PostCard(
+      title: "TCP Throughput",
+      category: .video,
+      description: "Learn all the things.",
+      backgroundColor: "green",
+      imageURL: "/static/images/workshop.jpg"
     )
   ]
 
--- a/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
+++ b/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
@@ -8,9 +8,7 @@ import SwiftyJSON
 
 struct LandingPageController {
 
-  static let eventPostCards : [PostCard] = EventServer.getEvents().flatMap { event in
-    return event.toPostCard()
-  }
+  static let eventPostCards : [PostCard] = EventServer.getEvents().flatMap{ $0.toPostCard() }
 
   static let videos = [
     PostCard(
@@ -46,7 +44,7 @@ extension Event {
           title: self.title,
           category: .workshop,
           description: self.eventDescription,
-          backgroundColor: self.colour, //TODO
+          backgroundColor: self.color, //TODO
           imageURL: self.imageURL
         )
       case is HackathonEvent:
@@ -54,7 +52,7 @@ extension Event {
           title: self.title,
           category: .hackathon,
           description: self.eventDescription,
-          backgroundColor: self.colour, //TODO
+          backgroundColor: self.color, //TODO
           imageURL: self.imageURL
         )
       default:
--- a/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
+++ b/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
@@ -8,7 +8,7 @@ import SwiftyJSON
 
 struct LandingPageController {
 
-  static let eventPostCards : [PostCard] = EventManager.events.flatMap { event in
+  static let eventPostCards : [PostCard] = EventServer.getEvents().flatMap { event in
     return event.toPostCard()
   }
 
--- a/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
+++ b/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
@@ -9,7 +9,7 @@ import SwiftyJSON
 struct LandingPageController {
 
   static let eventPostCards : [PostCard] = EventServer.getEvents().flatMap{ event in 
-    (event as? PostCardRepresentable)?.postCardRepresentation
+    event.postCardRepresentation
   }
 
   static let videos = [
--- a/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
+++ b/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
@@ -8,7 +8,9 @@ import SwiftyJSON
 
 struct LandingPageController {
 
-  static let eventPostCards : [PostCard] = EventServer.getEvents().flatMap{ $0.toPostCard() }
+  static let eventPostCards : [PostCard] = EventServer.getEvents().flatMap{ event in 
+    (event as? PostCardRepresentable)?.postCardRepresentation
+  }
 
   static let videos = [
     PostCard(
@@ -33,31 +35,5 @@ struct LandingPageController {
     } catch {
       Log.error("Socket error occured")
     }
-  }
-}
-
-extension Event {
-  func toPostCard() -> PostCard? {
-    switch self {
-      case is WorkshopEvent:
-        return PostCard(
-          title: self.title,
-          category: .workshop,
-          description: self.eventDescription,
-          backgroundColor: self.color, //TODO
-          imageURL: self.imageURL
-        )
-      case is HackathonEvent:
-        return PostCard(
-          title: self.title,
-          category: .hackathon,
-          description: self.eventDescription,
-          backgroundColor: self.color, //TODO
-          imageURL: self.imageURL
-        )
-      default:
-        return nil
-    }
-    
   }
 }
--- a/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
+++ b/Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
@@ -51,3 +51,29 @@ struct LandingPageController {
     }
   }
 }
+
+extension Event {
+  func toPostCard() -> PostCard? {
+    switch self {
+      case is WorkshopEvent:
+        return PostCard(
+          title: self.title,
+          category: .workshop,
+          description: self.eventDescription,
+          backgroundColor: self.colour, //TODO
+          imageURL: self.imageURL
+        )
+      case is HackathonEvent:
+        return PostCard(
+          title: self.title,
+          category: .hackathon,
+          description: self.eventDescription,
+          backgroundColor: self.colour, //TODO
+          imageURL: self.imageURL
+        )
+      default:
+        return nil
+    }
+    
+  }
+}
--- a/Sources/HaCWebsiteLib/Events/Event.swift
+++ b/Sources/HaCWebsiteLib/Events/Event.swift
@@ -8,6 +8,6 @@ protocol Event {
     var hypePeriod : DateInterval {get}
     var imageURL : String? {get}
     var location : Location? {get} //for room names etc.
-    var facebookID : Double? {get}
+    var facebookEventID : Double? {get}
     var shouldShowAsUpdate : Bool {get}
 }
--- a/Sources/HaCWebsiteLib/Events/Event.swift
+++ b/Sources/HaCWebsiteLib/Events/Event.swift
@@ -6,6 +6,7 @@ protocol Event: PostCardRepresentable {
     var tagLine : String {get}
     var color : String {get}
     var hypePeriod : DateInterval {get}
+    var tags : [String] {get}
     var imageURL : String? {get}
     var location : Location? {get}
     var facebookEventID : Double? {get}
--- a/Sources/HaCWebsiteLib/Events/Event.swift
+++ b/Sources/HaCWebsiteLib/Events/Event.swift
@@ -5,10 +5,10 @@ protocol Event {
     var time : Date {get}
     var eventDescription: String {get} //NOT related to event content (e.g. workshop content)
     var colour : String {get}
-    var hypeDays : Int {get}
-    var ttlDays : Int {get}
+    var hypePeriod : DateInterval {get}
     var imageURL : String? {get set}
     var endTime : Date? {get set}
     var venue : String? {get set} //for room names etc.
     var facebookLink : String? {get set}
+    var isLive : Bool {get}
 }
--- a/Sources/HaCWebsiteLib/Events/Event.swift
+++ b/Sources/HaCWebsiteLib/Events/Event.swift
@@ -4,11 +4,11 @@ protocol Event {
     var title : String {get}
     var time : Date {get}
     var eventDescription: String {get} //NOT related to event content (e.g. workshop content)
-    var colour : String {get}
+    var color : String {get}
     var hypePeriod : DateInterval {get}
     var imageURL : String? {get set}
     var endTime : Date? {get set}
     var venue : String? {get set} //for room names etc.
     var facebookLink : String? {get set}
-    var isLive : Bool {get}
+    var shouldShowAsUpdate : Bool {get}
 }
--- a/Sources/HaCWebsiteLib/Events/Event.swift
+++ b/Sources/HaCWebsiteLib/Events/Event.swift
@@ -9,6 +9,6 @@ protocol Event: PostCardRepresentable {
     var tags : [String] {get}
     var imageURL : String? {get}
     var location : Location? {get}
-    var facebookEventID : Double? {get}
+    var facebookEventID : String? {get}
     var shouldShowAsUpdate : Bool {get}
 }
--- a/Sources/HaCWebsiteLib/Events/Event.swift
+++ b/Sources/HaCWebsiteLib/Events/Event.swift
@@ -1,6 +1,6 @@
 import Foundation
 
-protocol Event {
+protocol Event: PostCardRepresentable {
     var title : String {get}
     var time : DateInterval {get}
     var tagLine : String {get}
--- a/Sources/HaCWebsiteLib/Events/Event.swift
+++ b/Sources/HaCWebsiteLib/Events/Event.swift
@@ -3,11 +3,11 @@ import Foundation
 protocol Event {
     var title : String {get}
     var time : DateInterval {get}
-    var eventDescription: String {get} //NOT related to event content (e.g. workshop content)
+    var tagLine : String {get}
     var color : String {get}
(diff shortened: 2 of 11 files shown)


Sources/HaCWebsiteLib/Events/EventManager.swift
```diff
@@ -0,0 +1,44 @@
+import Foundation
+
+public struct EventManager {
```
Pinpickle: Do we need an event manager?

Ideally we keep this website as stateless as possible, just something to interact with other stateful sources of truth (Database, workshop repo, etc...). In the case where there are expensive operations (such as pulling a Git repo), some layer of caching may be called for as an optimisation (ie `WorkshopManager`). In the case of _events_, we'll be pulling them from our database, which is not an expensive operation.

Do you think it would be less complex if we had an `events` computed property somewhere that does the logic you have in `update()` on every call?

Sources/HaCWebsiteLib/Events/Event.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+protocol Event {
+    var title : String {get}
+    var time : Date {get}
+    var eventDescription: String {get} //NOT related to event content (e.g. workshop content)
+    var colour : String {get}
+    var hypeDays : Int {get}
+    var ttlDays : Int {get}
```
Pinpickle: Thinking about the best way to do this one, but would the range between `hypeDays` and `ttlDays` be better expressed as a [DateInterval](https://developer.apple.com/documentation/foundation/dateinterval)?

Sources/HaCWebsiteLib/Events/HackathonEvent.swift
```diff
@@ -0,0 +1,27 @@
...
+
+struct HackathonEvent : Event {
+    let title : String
+    let time : Date 
+    let eventDescription: String
+    let colour : String
+    let hypeDays : Int
+    let ttlDays : Int
+    let duration : Int
+    var sponsor : String?
```
Pinpickle: Do these need to be mutable?

Sources/HaCWebsiteLib/Events/Event.swift
CharlieCrisp: Had no idea this datatype existed - that's going to make it soooo much nicer ;)

Sources/HaCWebsiteLib/Events/HackathonEvent.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
+
+struct HackathonEvent : Event {
```
Pinpickle: I like that we're creating different event types here. I think it would be valuable for us to figure out what these should be in advance.

Sources/HaCWebsiteLib/Events/HackathonEvent.swift
CharlieCrisp: So they were left mutable so that they could be optionally set but not in the constructor. I wasn't sure if it was possible to do this with immutables. I could add them to the constructor but I was worried about flexibility??

Sources/HaCWebsiteLib/Events/Event.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+protocol Event {
+    var title : String {get}
+    var time : Date {get}
+    var eventDescription: String {get} //NOT related to event content (e.g. workshop content)
+    var colour : String {get}
```
Pinpickle: Nit: very sorry to say that we're using American English in this repo (so `color` 😢)

Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
```diff
@@ -50,4 +36,30 @@ struct LandingPageController {
       Log.error("Socket error occured")
     }
   }
+}
+
+extension Event {
+  func toPostCard() -> PostCard? {
```
Pinpickle: Awesome that we're thinking ahead in how these events become post cards

Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
```diff
@@ -50,4 +36,30 @@ struct LandingPageController {
...
+      case is HackathonEvent:
+        return PostCard(
+          title: self.title,
+          category: .hackathon,
+          description: self.eventDescription,
+          backgroundColor: self.colour, //TODO
+          imageURL: self.imageURL
+        )
+      default:
+        return nil
```
Pinpickle: :+1: as this allows us (for whatever reason) to have events that don't make sense to be rendered in a post card.

Sources/HaCWebsiteLib/Events/Event.swift
CharlieCrisp: Damn - I guess it was just wishful thinking 😢

Pinpickle: Not a blocker but is there a type we can use to better represent a colour as well? Aware that we're using strings elsewhere but we could nip these magic strings in the bud before they get everywhere. Something like [UIKit's `UIColor`](https://developer.apple.com/documentation/uikit/uicolor)

Sources/HaCWebsiteLib/Events/Event.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+protocol Event {
+    var title : String {get}
+    var time : Date {get}
+    var eventDescription: String {get} //NOT related to event content (e.g. workshop content)
```
Pinpickle: I find the naming/documentation of this variable a tad confusing. On first read, I thought that the "workshop content" was suitable for this but anything to do with the event wasn't.

As it is the responsibility of `Event` implementations to provide this, is this detail something that's more important for `WorkshopEvent`?

I'm not sure if a `String` can represent the information required here (e.g. rich text w/ markdown). Workshops use this [`Text`](https://github.com/hackersatcambridge/hac-website/blob/master/Sources/HaCWebsiteLib/Text.swift) type.

Finally, how long is this description meant to be? `PostCards` require a few sentences, but an event page will likely have more detailed text. Do we need multiple descriptions to handle these cases?

Sources/HaCWebsiteLib/Events/Event.swift
```diff
@@ -0,0 +1,14 @@
...
+    var colour : String {get}
+    var hypeDays : Int {get}
+    var ttlDays : Int {get}
+    var imageURL : String? {get set}
```
Pinpickle: I see now why the variables in the implementations are mutable.

Why do these have to be mutable?

Sources/HaCWebsiteLib/Events/EventManager.swift
```diff
@@ -0,0 +1,44 @@
...
+
+        updatedEvents += [HackathonEvent(called: "Game Gig 2017", at: Date(), described: "Fun and Games", 
+            coloured: "black", hypePeriod: 7, coolOffPeriod: 7, lasting: 3)]
+
+        events = updatedEvents
+    }
+}
+
+extension Event {
+    func isLive() -> Bool {
```
Pinpickle: Nit: It's generally more Swifty to have computed properties over functions that don't have arguments (as long as the complexity is `O(1)`)

Sources/HaCWebsiteLib/Events/EventManager.swift
```diff
@@ -0,0 +1,44 @@
...
+        updatedEvents += workshopEvents
+
+        updatedEvents += [HackathonEvent(called: "Game Gig 2017", at: Date(), described: "Fun and Games", 
+            coloured: "black", hypePeriod: 7, coolOffPeriod: 7, lasting: 3)]
+
+        events = updatedEvents
+    }
+}
+
+extension Event {
```
Pinpickle: Nice use of extensions here. Do we want these to be private so they don't leak out into the rest of the project?

Sources/HaCWebsiteLib/Events/EventManager.swift
Pinpickle: Another note. I feel like `isLive` doesn't communicate what we're doing here. To me, it says "is this event going ahead or not, have we launched, is it published?". I imagine it is still possible to link to events even if this `isLive()` returns false.

Possibly `shouldShowAsUpdate` is a better name?

Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
```diff
@@ -7,37 +7,23 @@ import DotEnv
...
-  static let updates = [
-    PostCard(
-      title: "Conquering Linux",
-      category: .workshop,
-      description: "Be great at computers, not because you can, but because you dream.",
-      backgroundColor: "purple",
-      imageURL: "/static/images/workshop.jpg"
-    ),
+
+  static let eventPostCards : [PostCard] = EventManager.events.flatMap { event in
```
Pinpickle: 😍 `flatMap` to get rid of `nil`s

Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
```diff
@@ -7,37 +7,23 @@ import DotEnv
...
+    return event.toPostCard()
```
Pinpickle: Nit: Return is superfluous here. If you're feeling really adventurous, this entire closure can be written as `{ $0.toPostCard() }` (it varies from case to case whether that is more readable)

Sources/HaCWebsiteLib/Events/Event.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+protocol Event {
+    var title : String {get}
+    var time : Date {get}
```
Pinpickle: Can we also change `time` and `endTime` to a `DateInterval`?

Sources/HaCWebsiteLib/Events/Event.swift
```diff
@@ -0,0 +1,14 @@
...
+    var eventDescription: String {get} //NOT related to event content (e.g. workshop content)
+    var color : String {get}
+    var hypePeriod : DateInterval {get}
+    var imageURL : String? {get}
+    var endTime : Date? {get}
+    var venue : String? {get} //for room names etc.
```
Pinpickle: Did we want to have a location as coordinates here? We could encode this like

```
{
  coordinates: {
    latitude: Double;
    longitude: Double;
  }
  address?: String;
  venue?: String;
}

File: Sources/HaCWebsiteLib/Events/Event.swift [hunk 16]
```diff
@@ -0,0 +1,14 @@
...
+    var facebookLink : String? {get}
```
Pinpickle: What do we think about making this an ID instead of a URL? Allows us to generate the URL on the fly and potentially use the FB API if we want to.

Sources/HaCWebsiteLib/Events/Event.swift
CharlieCrisp: Yeah, good idea - do we think this will solely be used for Events? I.e. should I put it in the same directory as the rest of the events stuff?

Sources/HaCWebsiteLib/Events/Event.swift
CharlieCrisp: Cool - this takes out the duration property on `HackathonEvent` too :)

Sources/HaCWebsiteLib/Events/Event.swift
CharlieCrisp: As in this would be specifically the facebook event id? I'd be all for that :)

Sources/HaCWebsiteLib/Events/Event.swift
Pinpickle: I think this will be a useful struct for general use (for example rendering a location)

Sources/HaCWebsiteLib/Events/HackathonEvent.swift
CharlieCrisp: So we may need:
* HaC Nights
* Reading Groups
* Open Meetings 
* Q&As 
* ??
I don't necessarily think that all of these need their own types but maybe we should create a default class?

jaredkhan: What is the justification for these having separate types?

Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
jaredkhan: Might it be useful to have a protocol something like `PostCardRepresentable` which exposes a `postCardRepresentation` var?

Sources/HaCWebsiteLib/Events/HackathonEvent.swift
CharlieCrisp: Well we have different types for Workshops and Hackathons at the moment because each carries different information. I'm not sure there's a huge amount of justification for creating many more types beyond this though.

CharlieCrisp: Well, we have different types for Workshops and Hackathons at the moment because each carries different information. I'm not sure there's a huge amount of justification for creating many more types beyond this though.

Sources/HaCWebsiteLib/Events/Event.swift
CharlieCrisp: @Pinpickle So I think here what we could have is a `tagline` which would simply be a `String`. 

I think any sort of other `Description` probably needs to go in the implementation because not all events may need this. E.g. a hackathon would have its own separately developed page rather than a generated page (I imagine) and therefore I can't see where we would actually surface this description??

Sources/HaCWebsiteLib/Events/HackathonEvent.swift
Pinpickle: Maybe we couple the names slightly more to the data they hold so we don't need so many representations?

- `GeneralEvent` (for an event with a stock-standard page. Could fit open meetings, reading groups, Q&As, talks, etc...)
- `WorkshopEvent` (event linked to a workshop. Maybe a special case of `GeneralEvent`)
- `ExternalWebsiteEvent` (event linked to an external website such as Hack Cambridge)

If we weren't storing these all in a database, I'd be in favour of going crazy adding new types. Unfortunately these are likely all going to be in one DB table, so we may have to be a little pragmatic.

Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
Pinpickle: I like the way you think

Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
```diff
@@ -7,14 +7,15 @@ import DotEnv
...
-    PostCard(
-      title: "Conquering Linux",
-      category: .workshop,
-      description: "Be great at computers, not because you can, but because you dream.",
-      backgroundColor: "purple",
-      imageURL: "/static/images/workshop.jpg"
-    ),
+
+  static let eventPostCards : [PostCard] = EventServer.getEvents().flatMap{ event in 
+    guard let post = event as? PostCardRepresentable else {
```
CharlieCrisp: I think this does the right thing but I would appreciate a swifty person looking over this :wink:

Sources/HaCWebsiteLib/ViewModels/PostCard.swift
```diff
@@ -30,3 +31,7 @@ struct PostCard {
     )
   }
 }
+
+protocol PostCardRepresentable {
```
CharlieCrisp: @jaredkhan is this what you had in mind?

Sources/HaCWebsiteLib/Events/Event.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+protocol Event {
+    var title : String {get}
+    var time : DateInterval {get}
+    var tagLine : String {get}
```
CharlieCrisp: @Pinpickle here I've left out the `eventDescription` from the protocol but I've put it in some implementations where it's needed instead. I have included a `tagLine` instead.

Sources/HaCWebsiteLib/Events/ExternalWebsiteEvent.swift
```diff
@@ -0,0 +1,38 @@
...
+    let facebookEventID : Double?
+    var shouldShowAsUpdate: Bool {
+        get {
+            return self.hypePeriod.contains(Date())
+        }
+    }
+    var postCardRepresentation : PostCard {
+        return PostCard(
+          title: self.title,
+          category: .general,
```
CharlieCrisp: Currently, we're using `.general` as a `PostCategory` for all events - this is why I'd like to replace the `enum` with a member.

Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
jaredkhan: Could use chaining here to make the intent clearer:
(event as? PostCardRepresentable)?.postCardRepresentation

Sources/HaCWebsiteLib/Events/Event.swift
```diff
@@ -0,0 +1,13 @@
...
+    var color : String {get}
+    var hypePeriod : DateInterval {get}
+    var imageURL : String? {get}
+    var location : Location? {get} //for room names etc.
+    var facebookEventID : Double? {get}
+    var shouldShowAsUpdate : Bool {get}
```
jaredkhan: Perhaps we want events to have tags so we can implement a richer search later on (might be a lot of work to add tags retrospectively)

Sources/HaCWebsiteLib/Events/ExternalWebsiteEvent.swift
```diff
@@ -0,0 +1,38 @@
...
+          description: self.tagLine,
+          backgroundColor: self.color, //TODO
+          imageURL: self.imageURL
+        )
+    }
+
+    init(called title: String, at time: DateInterval, summarised tagLine : String, colored color: String,
```
jaredkhan: Often different argument labels don’t add much to the clarity of the API at the point of use. Here, the user of the API doesn’t necessarily know that `at: ` should be followed by a time. Using the parameter name `time: ` may make this clearer for the user. The same goes for (arguably) the rest of the parameters.

Sources/HaCWebsiteLib/ViewModels/PostCard.swift
jaredkhan: Looks good!

Sources/HaCWebsiteLib/Events/ExternalWebsiteEvent.swift
CharlieCrisp: You're right - I was just trying too hard and it just made things worse lol ;)

Sources/HaCWebsiteLib/Events/GeneralEvent.swift
```diff
@@ -0,0 +1,44 @@
...
+    let title : String
+    let time : DateInterval
+    let tagLine : String
+    let color : String
+    let hypePeriod : DateInterval
+    let tags : [String]
+    let eventDescription : Text
+    let imageURL : String? 
+    let location : Location?  //for room names etc.
+    let facebookEventID : Double?
```
Pinpickle: How come the event is a `Double`? Do you think `String?` would be a better type?

CharlieCrisp: I put it as a `Double` because I found somewhere that Facebook uses a 64 bit ID. I guess wrt `Strings` I'm not really sure what the benefit of one or the other would be.

Pinpickle: I worry about `Double` because we can't really get away with floating-point precision when we're trying to reference an exact thing.

The [Facebook API reference](https://developers.facebook.com/docs/graph-api/reference/event/) says that IDs are a "numeric string", so while they are a number, they come back as a string. We can treat them as black boxes in this case (I believe Facebook expects us to), so if IDs ever switch to something else (like a UUID for example), we won't get hit by it.

Sources/HaCWebsiteLib/Events/Event.swift
```diff
@@ -0,0 +1,14 @@
...
+
+protocol Event: PostCardRepresentable {
+    var title : String {get}
+    var time : DateInterval {get}
+    var tagLine : String {get}
+    var color : String {get}
+    var hypePeriod : DateInterval {get}
+    var tags : [String] {get}
+    var imageURL : String? {get}
+    var location : Location? {get} //for room names etc.
```
Pinpickle: I can't seem to find this `Location` type anywhere in Foundation or our repo. Could you tell me where this is coming from?

CharlieCrisp: I think it's just in the HaCWebsiteLib directory

Pinpickle: GitHub was weird and I missed the file, let's carry on.

Is the comment still necessary here?

CharlieCrisp: nope it's not necessary  -will fix now

Sources/HaCWebsiteLib/Events/ExternalWebsiteEvent.swift
```diff
@@ -0,0 +1,40 @@
+import Foundation
+
+struct ExternalWebsiteEvent : Event {
```
Pinpickle: As this is supposed to be referring to an external website, should it have a URL pointing to said external website?

jaredkhan:
*Retrospective approve* Lovely stuff Charlie! 👌

PR #57, opened 2017-09-18
Add SwiftLint support

Pinpickle:
This PR adds support for [SwiftLint](https://github.com/realm/SwiftLint), configures some rules that were particularly grating, and fixes all of the offenders in our current codebase.

Caveats:

- Fixing is only partially automatic. There is a `swiftlint autocorrect` feature but will leave many rules for manual fixing
- Have not got it configured with Docker. Linux support exists but takes some love. So for this time, it's Mac only and not integrated into our CI builds.

I also did a teeny weeny bit of refactoring (only a little bit).

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,10 @@
+included:
+  - Sources
+excluded:
+  - .build
+disabled_rules:
+  - todo
+type_name:
+  min_length: 1
+identifier_name:
+  min_length: 1
--- a/Sources/HaCWebsiteLib/ViewModels/LandingPage.swift
+++ b/Sources/HaCWebsiteLib/ViewModels/LandingPage.swift
@@ -1,5 +1,7 @@
 import HaCTML
 
+// swiftlint:disable line_length
+
 struct LandingPage {
   let updates: [PostCard]
 
--- a/Sources/HaCWebsiteLib/routes.swift
+++ b/Sources/HaCWebsiteLib/routes.swift
@@ -13,23 +13,16 @@ func getWebsiteRouter() -> Router {
 
   router.get("/", handler: LandingPageController.handler)
 
-  router.get("/workshops") { request, response, next in
-    defer {
-      next()
-    }
-    do {
-      try response.send(
-        UI.Pages.workshops(workshops: WorkshopManager.workshops).render()
-      )
-      .end()
-    } catch {
-      Log.error("Failed to render template \(error)")
-    }
+  router.get("/workshops") { _, response, next in
+    try response.send(
+      UI.Pages.workshops(workshops: WorkshopManager.workshops).render()
+    ).end()
+    next()
   }
 
   /// Intended for use by GitHub webhooks
-  router.post("/api/refresh_workshops") { request, response, _ in
-    try! WorkshopManager.update()
+  router.post("/api/refresh_workshops") { _, _, _ in
+    try WorkshopManager.update()
   }
 
   return router
--- a/Sources/HaCTML/Attribute.swift
+++ b/Sources/HaCTML/Attribute.swift
@@ -32,12 +32,13 @@ public struct AttributeKey<Value> {
     self.keyName = key
     self.applyFunc = apply
   }
-  
+
   internal func apply(_ value: Value) -> AppliedAttribute {
     return (keyName: keyName, value: applyFunc(value))
   }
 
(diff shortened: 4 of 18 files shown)


Sources/HaCWebsiteLib/routes.swift
```diff
@@ -13,23 +13,16 @@ func getWebsiteRouter() -> Router {
 
   router.get("/", handler: LandingPageController.handler)
 
-  router.get("/workshops") { request, response, next in
-    defer {
-      next()
-    }
-    do {
```
Pinpickle: Turns out errors are automatically caught by Kitura: https://github.com/IBM-Swift/Kitura/commit/42eb6bf947f7ae49e22e16deedaad3935217729c

This means we can catch these errors and do something with them in a general-purpose error handling middleware. This is slated for #58

Sources/HaCWebsiteLib/routes.swift
```diff
@@ -13,23 +13,16 @@ func getWebsiteRouter() -> Router {
...
+  router.get("/workshops") { _, response, next in
+    try response.send(
+      UI.Pages.workshops(workshops: WorkshopManager.workshops).render()
+    ).end()
+    next()
   }
 
   /// Intended for use by GitHub webhooks
-  router.post("/api/refresh_workshops") { request, response, _ in
-    try! WorkshopManager.update()
```
Pinpickle: Willing to discuss removing the force try here, in which case we can leave it out of the PR and come back to the idea later.

CharlieCrisp: So am I right in thinking that with the force try, it would crash if `WorkshopManager` crashes, and without it, it would use a potentially out of date set of workshops?

Pinpickle: Correct. We can add logging here and/or tests on the workshop manager repo to catch this stuff. I'm all for failing fast, but in this case an update to a GitHub repo can crash our app automatically. That's some spooky action at a distance right there.

jaredkhan: 👍

Sources/HaCWebsiteLib/routes.swift
jaredkhan: Much nicer! 👍 as long as we're logging the Error

Pinpickle: Alas, we aren't yet. That's what the referenced issue (#58) is for. Is not logging a blocker?

jaredkhan: Nah, the issue is tagged appropriately so we Gucci

Sources/HaCWebsiteLib/ViewModels/LandingPage.swift
```diff
@@ -1,5 +1,7 @@
 import HaCTML
 
+// swiftlint:disable line_length
```
jaredkhan: Seems like we might have to disable this for most ViewModels 🤷‍♀️

Pinpickle:
@jaredkhan no limitations on the Linux setup, just takes some time for us to put the docker config together (need to install SourceKit, set up build environment for SwiftLint, add it to path), then our Windows users can use it.

Definitely something we want to do, just testing the waters atm.

Sources/HaCWebsiteLib/ViewModels/LandingPage.swift
Pinpickle: Yeah I think so

Alternatively we can try to keep our line lengths shorter ;)

.swiftlint.yml
```diff
@@ -0,0 +1,10 @@
+included:
+  - Sources
+excluded:
+  - .build
+disabled_rules:
```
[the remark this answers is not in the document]
Pinpickle: I'm happy to go down that direction, but I trust that SwiftLint have sensible defaults here. I'd much rather run into frustration for an overly restrictive rule and disable it, as opposed to missing out on a potentially useful rule.

Coming from JS linting land, maintaining a blacklist was much easier than maintaining a whitelist.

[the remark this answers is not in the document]
jaredkhan: That's fair. I suppose with pinning we're also not really in danger of new rules sneaking in without action from us. Let's stick to blacklist.

jaredkhan:
Let's wait until Charlie's changes go in to merge this

PR #60, opened 2017-09-29
Fix warning during `docker-compose build`

moosichu:
```bash
[WARNING]: Empty continuation line found in:
    RUN   export DEBIAN_FRONTEND=noninteractive &&   apt-get update &&   apt-get install curl apt-transport-https &&   curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash - &&   curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - &&   echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list &&   apt-get update &&   apt-get install nodejs yarn &&   npm install -g n &&   n 6.11.3
[WARNING]: Empty continuation lines will become errors in a future release.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,13 @@
 FROM ibmcom/swift-ubuntu:3.1
 
+# Allow installation of Node 6: https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions \
+# Install Yarn: https://yarnpkg.com/en/docs/install#linux-tab
+
 RUN \
   export DEBIAN_FRONTEND=noninteractive && \
   apt-get update && \
   apt-get install curl apt-transport-https && \
-  # Allow installation of Node 6: https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions \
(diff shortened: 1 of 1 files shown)


moosichu:
@jaredkhan I don't think so, as I think they are referencing where we got our information on how to install the relevant libraries - having them at the head of the Dockerfile isn't really out of place.

jaredkhan:
Fair but it's also nice having them next to the commands they relate to. Is the solution here to just have lots of `RUN` lines?

Pinpickle:
`RUN` lines are problematic as they introduce new layers in our image (making it bigger).

The ideal way forward is to use lots of run commands in [multistage builds](https://docs.docker.com/engine/userguide/eng-image/multistage-build/)

moosichu:
Shall we park this pull request and make this an issue then? It seems like the solution might be a bit more involved.

jaredkhan:
RE @Pinpickle's comment:
Good post about 'layers' [here](https://medium.com/@jessgreb01/digging-into-docker-layers-c22f948ed612). *TL;DR* the effect of each command in a Dockerfile is stored as a separate diff and cached so that it can be applied to other Docker images

PR #62, opened 2017-10-02
Remove empty continuation lines in Dockerfile

moosichu:


jaredkhan:
Could you explain why this was needed?

moosichu:
See pull request #60 , empty continuation lines will be deprecated in future versions of docker.

Further reading for background info: https://github.com/moby/moby/issues/29005

Pinpickle:
Not super problematic here though as we're not really doing any cleanup in these run commands - splitting them up shouldn't have too much overhead. I'm not that clued-up on Docker internals though.

PR #67, opened 2017-10-02
Content/about section

moosichu:
It's not in a final and ideal state yet - but should be good enough for tomorrow.


Sources/HaCWebsiteLib/ViewModels/LandingPage.swift
```diff
@@ -4,6 +4,12 @@ import HaCTML
 
 struct LandingPage {
   let updates: [PostCard]
+  let youtube_url = "https://www.youtube.com/hackersatcambridge"
```
Pinpickle: Nit: we're camelCase here (so `youtubeURL`)

moosichu: Fixing now...

moosichu: foxed @_

PR #70, opened 2017-10-02
Feature/mvp events

moosichu:
Removes the stream of updates for the purposes of MVP and makes some minor fixes, including linking `find out more` to the about section.


moosichu:
Also includes changes from #68, so merge that first.

Issue #72, opened 2017-10-03
Re-add the event stream


moosichu:
Got removed for MVP, but should be re-added in a polished way.

PR #75, opened 2017-10-04
Create a route `beta/event-stream` for developing the event stream

moosichu:
Resolves issue #72

### DIFF
--- a/Sources/HaCWebsiteLib/ViewModels/EventStream.swift
+++ b/Sources/HaCWebsiteLib/ViewModels/EventStream.swift
@@ -1,0 +1,22 @@
+import HaCTML
+
+// swiftlint:disable line_length
+
+struct EventStream {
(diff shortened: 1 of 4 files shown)


Sources/HaCWebsiteLib/ViewModels/EventStream.swift
```diff
@@ -0,0 +1,22 @@
+import HaCTML
+
+// swiftlint:disable line_length
+
+struct EventStream {
```
jaredkhan: This is more accurately named "LandingUpdateFeed". If we'd like to make this general-purpose for anywhere we use an Update Feed, we'll want to strip the word `Landing` from the CSS classes below and in the stylesheets

moosichu:
@jaredkhan - I've implemented the changed, could you please review again?

PR #79, opened 2017-10-07
Improve styling for squash feature

Pinpickle:
image


Package.resolved
```diff
@@ -0,0 +1,214 @@
+{
```
CharlieCrisp: @Pinpickle what is this file? It looks like it's been generated but not filtered out by the gitignore...

Pinpickle: Whoops I thought I deleted this. I think it's a Swift 4 thing that I accidentally created while debugging.

Issue #78, opened 2017-10-06
Create a nice 404 page for when an incorrect route is entered


moosichu:
See issue #58 for the backend needed for this.

When someone goes to [https://hackersatcambridge.com/hot-hot-hotdogs](https://hackersatcambridge.com/hot-hot-hotdogs) (or any other invalid route), you get something similar to:

![Screen Of Webpage Showing "Cannot GET /hot-hot-hotdogs."](https://user-images.githubusercontent.com/10348641/31521345-0f67a504-afa1-11e7-967c-be948e58458b.png)

Which is naturally incredibly ugly and unhelpful, with no way of navigating to a valid page.

Ideally we want a nicely themed 404 page. This issue is purely for creating the visuals of such a page, #58 is about the backend system to route to such a page.

moosichu:
@Pinpickle Is this actually a good first issue, and if so what is the best place to get started?

Pinpickle:
@moosichu firstly this is a dupe of #58, but it probably benefits from being split up - we'd just need to edit the original issue.

This one would require ramping up of both templating (rendering the 404 page) and routing (catching the unfound route). Maybe we should split this up into the template and the routing?

moosichu:
@Pinpickle Gonna make this the templating one then

Pinpickle:
Fixed by #86 (including the routing!)

Issue #58, opened 2017-09-18
Add error handling for routes (with error pages)


Pinpickle:
Should handle not found pages and errors thrown in routes.

Need a pretty page to tell our lovely users when something has gone wrong.

This issue is about the backend part of this, see issue #78 for the frontend.

jaredkhan:
Alright have spent too much time and am none-the-wiser on how to implement a simple 404 catcher 🙄
Unassigning myself for now and recommending we don't attempt to do this in time for the fair

PR #86, opened 2017-10-14
Add 404 page and supporting back end

doctorn:


doctorn:
This definitely resolves #78 but after reading #58 more thoroughly I'm not entirely sure it solves the issue of handling errors thrown in routes.

Sources/HaCWebsiteLib/ViewModels/NotFound.swift
```diff
@@ -0,0 +1,27 @@
...
+
+struct NotFound {
+  
+  var node: Node {
+    return UI.Pages.base(
+      title: "Page Not Found",
+      content: Fragment(
+        El.Div[Attr.className => "NotFound_container"].containing(
+          El.Img[
+            Attr.className => "LandingIntroduction__titleImage",
```
Pinpickle: So we're using a bunch of classes from the `LandingIntroduction` block here. Our conventions specify that they don't make sense outside of the context of their containing block, and may break at any time. Semantically, this is also questionable.

Essentially, people would eventually change (or even remove) this `LandingIntroduction` block, and it won't be obvious that it will affect our 404 page as well, possibly breaking it.

We have some options to get around this:

- Create elements specific to `NotFound` (e.g. `NotFound__logoImage`); _or_
- Abstract these elements into blocks of their own so they can actually be re-usable components.

For example, we could break this logo + "Cambridge's student tech society" thing into it's own block, `SiteLogo`, with a corresponding view model so we can reuse it.

Sources/HaCWebsiteLib/ViewModels/NotFound.swift
```diff
@@ -0,0 +1,27 @@
...
+            Attr.src => "/static/images/hac-logo-dark.svg",
+            Attr.alt => "Hackers at Cambridge"
+          ],
+          El.Div[Attr.className => "LandingIntroduction__tagLine"].containing("Cambridge's student tech society"),
+          El.H1[Attr.className => "NotFound_title"].containing("404"),
+          El.P.containing("We can't find the page you're looking for..."),
+          El.A[Attr.href => "/"].containing(
+            El.Div[Attr.className => "LandingIntroduction__callToActionButton"].containing(
```
Pinpickle: Similar to the comments above, would it make sense to break this out into, for example, a `BigButton` block?

static/src/styles/NotFound.styl
```diff
@@ -0,0 +1,16 @@
...
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  position: absolute;
+  left: 50%;
+  top: 50%;
+  transform: translate(-50%, -50%);
+}
+
+.NotFound_title {
```
Pinpickle: Nit: elements have _two_ underscores between block and element name (so `.NotFound__title`)

static/src/styles/NotFound.styl
```diff
@@ -0,0 +1,16 @@
...
+	font-size: 100px;
```
Pinpickle: Nit: Spacing (2 spaces instead of a tab)

jaredkhan:
@doctorn Thanks for this contribution! For frontend changes, please add screenshots to the comments (sorry this wasn't in the readme)

Sources/HaCWebsiteLib/NotFoundHandler.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+import Kitura
+
+public class NotFoundHandler: RouterMiddleware {
+
+  public func handle(request: RouterRequest, response: RouterResponse, next: @escaping () -> Void) {
+    defer {
+      next()
+    }
+    if response.statusCode != .OK {
```
Pinpickle: I pulled this down and had a check, errors thrown in routes _do not_ get caught by this, so we'll still need to work on fixing that somehow.

Those errors are much rarer though, so this solves the most pressing case.

Sources/HaCWebsiteLib/ViewModels/NotFound.swift
```diff
@@ -0,0 +1,27 @@
+import HaCTML
+
+struct NotFound {
+  
+  var node: Node {
+    return UI.Pages.base(
+      title: "Page Not Found",
+      content: Fragment(
+        El.Div[Attr.className => "NotFound_container"].containing(
```
Pinpickle: I am so glad to see you using HaCTML here very fluently. If you ran into anything confusing on the journey, could you let us know?

README.md
```diff
@@ -35,7 +35,7 @@ When you want to run the project again (and you will, it's great):
 
 ### Environment Variables
 
-Your `.env` file should contain the following items in the format: _key_=_value_
+Your `.env` file should contain the following items in the format: _key_=_"value"_
```
Pinpickle: Contributing code _and_ fixing our docs, you're making the world a better place

moosichu: Why the quotes? I don't have them in my `.env` file.

static/src/styles/NotFound.styl
moosichu: @Pinpickle out of interest, why is it two vs one? It's not easy to tell the difference anyway. So I don't see how it helps?

Pinpickle: The underlying decisions are mostly arbitrary, but we chose it because it's a widely-adopted convention: https://github.com/airbnb/css#oocss-and-bem

It's much easier to say "we follow the Airbnb style guide" and let them justify everything, which they do quite well.

README.md
doctorn: It wouldn't build on my machine without them

doctorn:
Here is the screenshot

screen shot 2017-10-14 at 18 11 59

README.md
moosichu: What OS are you running?

moosichu: (Although it shouldn't matter as we are using containers)
@Pinpickle What do you do?

doctorn: MacOS at the moment as I couldn't get Swift to install on Linux

moosichu: Oh, are you running the project in docker?

doctorn: I'm using docker on my Mac, but I'll try later on my laptop

doctorn:
@Pinpickle have you had a chance to read over the new commit?

Sources/HaCWebsiteLib/ViewModels/NotFound.swift
```diff
@@ -0,0 +1,27 @@
...
+
+struct NotFound {
+  
+  var node: Node {
+    return UI.Pages.base(
+      title: "Page Not Found",
+      content: Fragment(
+        El.Div[Attr.className => "NotFound"].containing(
+          El.Img[
+            Attr.className => "LandingIntroduction__titleImage",
```
Pinpickle: It looks like we still have `LandingIntroduction__titleImage` and `LandingIntroduction__tagLine` here?

static/src/styles/core/interaction.styl
```diff
@@ -0,0 +1,13 @@
+.Interaction__bigButton {
```
Pinpickle: Great work abstracting these out into reusable components 😄 

_Tiny_ nit here. `bigButton` is technically a block, not an element (but is being named as an element in this case). It seems like you're using the block segment (`Interaction` in this case) as a namespace. I appreciate the need for namespaces, and maybe we do need them, so in this case, `BigButton` or `InteractionBigButton` would match our styleguide.

Sources/HaCWebsiteLib/ViewModels/NotFound.swift
doctorn: My thought was that the 404 page styling is dependent on how you style the logo and strapline in general, but I wasn't sure how you wanted that styling separated from the rest of the landing page styling.

Pinpickle: I agree - it makes sense that this construct is consistent where we use it on the site. This leads to our naming being incorrect in this case, if it's used in more places than our landing page.

Do you think that the same approach you used for the button would make sense here? Pulling it out into some kind of `SiteLogo` component?

doctorn: Ok yeah I'll do that now

Pinpickle:
Tests are passing, looks good to me, let's get this thing out there!

Great work @doctorn

doctorn:
Thanks ^-^

Pinpickle:
@doctorn your changes are now live! https://hackersatcambridge.com/definitely-not-a-page

PR #88, opened 2017-10-14
Add viewport meta tag

### LINKED ISSUE #81: Viewport meta tag is missing
This is required for correct scaling on mobile. More info [here](https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag)

CharlieCrisp:
This PR fixes #81 by adding a meta tag to the `base` `View`.
This PR also fixes a bug where the text and images on the landing page were misaligned.
Finally, this PR changes the last picture of the `LandingAbout` section so it's not the same as the first.

###  Viewport fix:
Previously on mobiles
image

Now on mobiles
image

### Misalignment bug:
Previously
image

Now
image

### Final Picture
Previously
image

Now
image


Sources/HaCTML/Attributes.swift
```diff
@@ -25,10 +25,12 @@ public enum Attributes {
   static public let alt = stringAttribute("alt")
   static public let charset = stringAttribute("charset")
   static public let className = stringAttribute("class")
+  static public let content = stringAttribute("content")
```
Pinpickle: Improving HaCTML _and_ keeping things in alphabetical order - nice and tidy 📐

static/src/styles/landing/LandingAbout.styl
```diff
@@ -1,17 +1,22 @@
...
+.LandingAbout__section {
   padding: grid(4);
 }
 
 .LandingAbout__imageContainer {
   flex: 1;
 }
 
+.LandingAbout__section:nth-child(even) .LandingAbout__imageContainer {
+  order: 10;
```
Pinpickle: While this isn't in effect in mobile views as display isn't flexbox, do you think it would better describe intent if this declaration were in the media query below?

At first when reading this, I was confused as I thought there would be alternating images in our mobile view - I had to think through it to figure out what was actually going on.

moosichu:
My only comment is that commit a17e72d4a8bc346111aa27d84b65912f7c011d66 has a commit comment that doesn't fit our guidelines. That's it. :stuck_out_tongue:. (And more of a comment for future commits and not really a deal breaker.)

static/src/styles/landing/LandingAbout.styl
CharlieCrisp: @Pinpickle I think you're right - should be fairly trivial to change too 😄

Issue #66, opened 2017-10-02
Make compile errors less silent


moosichu:
Currently, when a compile error occurs in an edited swift file, you have to investigate the log to see if such a thing actually occurred. It would be nice if the website itself displayed a message showing the compile error. (Or at least saying that an error had occurred.)

moosichu:
Closed by #90

Issue #74, opened 2017-10-04
Add a favicon.ico


moosichu:

moosichu:
@Pinpickle This could be a really good first issue if we provide the favicon.ico here.

moosichu:
Closed by #90

Issue #45, opened 2017-09-01
Shouldn't reload server when Swift build fails


jaredkhan:
Doing this loses the actual swift error in the noise.
To do this, we'll need our swiftBuild task in the gulpfile to emit a stream end, maybe like the styles build does

jaredkhan:
Struggling on this one. Can easily enough throw an error from the build task but the series currently doesn't catch the error so need to figure out how to do that

moosichu:
This seems related to #66, but the issues suggest alternate solutions to the same problem

PR #90, opened 2017-10-16
Add error output to website if 'swift build' fails

doctorn:
The output of `swift build` will now be displayed on all served web pages if build fails
This PR also adds the favicon to the base view model (someone will have to chose a better image for it though)


doctorn:
Closes #66 and #74 (with temporary favicon)

doctorn:
image

doctorn:
Also #45 is covered by this as the entire error log appears in the web page after the server is reloaded

PR #92, opened 2017-10-18
Add generic redirects middleware + intro redirect

Pinpickle:
This gives us the ability to make a list of static redirects from paths to arbitrary URL strings.

Included is a redirect to our "Intro to programming" workshop.


jaredkhan:
From discussion offline:
Let's move the list of redirects into `routes.swift` for colocation rather than having this defaultRedirects thing

PR #94, opened 2017-10-22
Update landing feature

jaredkhan:
Updates the landing feature and adds a little structure for future features, in preparation for adding a controller that will handle these changes for us.

Screenshots:
Previously
screen shot 2017-10-22 at 09 35 51

Now
screen shot 2017-10-22 at 09 34 40


jaredkhan:
@moosichu Not all features will necessarily be images. We might want to have animated graphics here for some larger events (e.g. Hack Cambridge). I think having the feature hero take arbitrary Nodes is the most flexible way forward. We can sugar up this image/background thing if we like.

PR #96, opened 2017-10-22
Move not found handler to later position

jaredkhan:
In rearranging some routes in a previous PR, I broke the whole site. Let's undo that.


doctorn:
Yeah for future reference the way the 404 handler is designed means that it has to be the last handler is checked (basically it only adds the 404 page to the response if no previously checked handler returned a 200 status code).

PR #98, opened 2017-10-24
Move middleware to its own folder

jaredkhan:


Pinpickle:
@jaredkhan you think we should rename `NotFoundHandler` to `NotFoundMiddleware` while we're doing this?

jaredkhan:
@Pinpickle Good call

jaredkhan:
@Pinpickle all yours

PR #100, opened 2017-10-25
Add Bash workshop feature

jaredkhan:
Will also change the link to the description when appropriate for less stress this evening!

Previously
screen shot 2017-10-25 at 14 57 11

Now
screen shot 2017-10-25 at 14 57 02


Sources/HaCWebsiteLib/ViewModels/LandingFeatures/BashWorkshop.swift
```diff
@@ -0,0 +1,30 @@
...
+import Foundation
+
+extension LandingFeatures {
+  struct BashWorkshop {
+    let facebookLink = "https://www.facebook.com/events/245447659318325"
+    let descriptionLink = "/bash"
+
+    /// Returns a link to the most currently relevant information about this workshop
+    var link: String {
+      // 2017-10-25 18:45
```
Pinpickle: In the future, let's try to find a better way to navigate around timezones so these comments aren't necessary.

PR #101, opened 2017-10-25
Give ImageHero an optional date field

Pinpickle:
Currently the workflow for adding features requires some poor soul to go and manually add a date to the image.

This PR adds a consistent way to add dates to the feature, so we don't have to keep doing that.

Also adds optional support to HaCTML, and touches up the Bash feature a bit.

Previously
image

Now
image


Sources/HaCWebsiteLib/ViewModels/LandingFeatures/BashWorkshop.swift
```diff
@@ -5,11 +5,11 @@ extension LandingFeatures {
   struct BashWorkshop {
     let facebookLink = "https://www.facebook.com/events/245447659318325"
     let descriptionLink = "/bash"
+    // 2017-10-25 18:45
+    let workshopStartDate = Date(timeIntervalSince1970: 1508953500)
```
CharlieCrisp: Could we maybe abstract this into a function which takes a year, month, day... and returns the relevant number? Does that function already exist?

Sources/HaCWebsiteLib/ViewModels/ImageHero.swift
```diff
@@ -18,12 +38,36 @@ struct ImageHero {
     case color(String)
   }
 
+  enum Color {
```
CharlieCrisp: I feel like the name of this is a bit misleading. I'm not entirely sure what would be better though, but currently, this deals with tones not colours if I've understood it right.

Sources/HaCWebsiteLib/ViewModels/ImageHero.swift
```diff
@@ -1,15 +1,35 @@
...
 
 struct ImageHero {
   let background: BackgroundType
   let imagePath: String
   let alternateText: String
   let destinationURL: String?
+  let dateToDisplay: Date?
+  let textColor: Color
+
+  init(
```
jaredkhan: Can still take advantage of default initialisers here, can put the default values up here

Sources/HaCWebsiteLib/ViewModels/ImageHero.swift
jaredkhan: Agreed this is a little unclear`BackgroundColourType` might be more meaningful

Sources/HaCWebsiteLib/ViewModels/ImageHero.swift
```diff
@@ -18,12 +38,36 @@ struct ImageHero {
     case color(String)
   }
 
+  enum Color {
+    case dark
+    case light
+  }
+
+  private var dateBlock: Node? {
+    return formattedDate.map({ dateString in El.Div[Attr.className => "ImageHero__date"].containing(dateString) })
```
jaredkhan: This is still a little unergonomic

Sources/HaCWebsiteLib/ViewModels/LandingFeatures/BashWorkshop.swift
jaredkhan: Am working on it

Sources/HaCWebsiteLib/ViewModels/ImageHero.swift
Pinpickle: We talking about HaCTML here or the `map` construct I'm using?

jaredkhan: Bit of both
Think optionals make sense in general for Nodes that may or may not exist but the mapping is not very intuitive so we should explore other options for this. Nathan brought this up too. Not something that needs to be addressed right now

Sources/HaCTML/Nodeable.swift
```diff
@@ -0,0 +1,18 @@
...
+  var node: Node { get } 
+}
+
+public extension Nodeable {
+  func containedBy(element: HTMLElement) -> Node {
+    return element.containing(self)
+  }
+}
+
+extension String: Nodeable {
```
jaredkhan: 😍

Issue #25, opened 2017-04-03
Way to control featured topic on homepage


jaredkhan:
When an event is currently happening, it should be featured. This should be automatic
We should also be able to control the featured topic text/link for when there is no event currently on. More automatic controls can be added later down the line.

PR #102, opened 2017-10-28
Add a way to schedule Features

jaredkhan:
- Addresses #25 by providing a simple date-based schedule for home page features and allowing features to expire after a certain date (showing no feature if there is nothing but old features available).

- Adds an EventFeature model for constructing these Features for events that have dates associated with them.

- Adds a way to write out dates without resorting to UNIX timestamps.

- Adds the Feature for the git workshop.

- Restructures so that the ImageHero component doesn't own the Date component or linking functionality (both now owned by the EventFeature component).

Previously
screen shot 2017-10-28 at 18 33 16

Now
screen shot 2017-10-28 at 18 36 11


Sources/HaCWebsiteLib/Utils/DateUtils.swift
```diff
@@ -0,0 +1,23 @@
...
+    return formatter
+  }()
+}
+
+extension Date {
+  static func from(year: Int, month: Int, day: Int, hour: Int, minute: Int, timezone: TimeZone? = nil) -> Date! {
+    let formatter = DateFormatter()
+    formatter.dateFormat = "yyyy-MM-dd HH:mm"
+    formatter.timeZone = timezone ?? TimeZone(identifier: "Europe/London")
+    return formatter.date(from: "\(year)-\(month)-\(day) \(hour):\(minute)")
```
Pinpickle: I'm flabbergasted that we have to work this hard to provide a _lower_ level representation of a date than what the std lib gives us but it's abstracted away!

People may claim that datestrings are easier to work with in this scenario but I really like the type safety this is giving us.

Sources/HaCWebsiteLib/ViewModels/ImageHero.swift
```diff
@@ -5,77 +5,28 @@ struct ImageHero: Nodeable {
...
-    return El.Div[Attr.className => className]
-  }
 
-  private var backgroundStyle: [String: String] {
-    switch (background) {
-    case .image(let path):
-      return ["background-image": "url('\(path)')"]
-    case .color(let color):
-      return ["background-color": color]
+    var style: [String: String] {
```
Pinpickle: This is so cool

Sources/HaCWebsiteLib/ViewModels/LandingPage/LandingFeatures/LandingFeatures.swift
```diff
@@ -0,0 +1,17 @@
...
+    LandingFeatures.introToProgramming1,
+    LandingFeatures.introToProgramming2,
+    LandingFeatures.bashWorkshop,
+    LandingFeatures.gitWorkshop
+  ]
+
+  /// Gets the most currently appropriate feature
+  static var currentFeature: LandingFeature? {
+    let currentDate = Date()
+    return features.first { $0.expiryDate > currentDate }
```
Pinpickle: So declarative

This is giving me chills

Tests/HaCWebsiteLibTests/DateUtilsTests.swift
```diff
@@ -0,0 +1,36 @@
+import XCTest
+@testable import HaCWebsiteLib
+
+class DateUtilsTests: HaCWebsiteLibTestCase {
```
Pinpickle: Tests as well? Daaaaaaamn.

Sources/HaCWebsiteLib/Utils/DateUtils.swift
```diff
@@ -0,0 +1,23 @@
...
+    formatter.dateStyle = .medium
+    formatter.timeStyle = .none
+    formatter.locale = locale
+
+    return formatter
+  }()
+}
+
+extension Date {
+  static func from(year: Int, month: Int, day: Int, hour: Int, minute: Int, timezone: TimeZone? = nil) -> Date! {
```
Pinpickle: What's the reasoning for using an implicitly unwrapped optional here?

Tests/HaCWebsiteLibTests/DateUtilsTests.swift
```diff
@@ -0,0 +1,36 @@
...
+  func testDate() {
+    let readableDate = Date.from(year: 2017, month: 10, day: 25, hour: 19, minute: 00)
+    let timestampDate = Date(timeIntervalSince1970: 1508954400)
+
+    XCTAssertEqual(readableDate, timestampDate)
+  }
+
+  func testTwiceOccurringDate() {
```
Pinpickle: I think these tests might be _slightly_ flakey in that if timezones get redefined for some reason, these dates can shift from UTC.

I don't know if that's allowed to happen for past dates :thinking_face: 

Not a blocker though

Sources/HaCWebsiteLib/Utils/DateUtils.swift
jaredkhan: Reasoning was that it'd be nice if we don't have to propagate optionals all over here. Since we're hard coding dates for now, we don't want the site to successfully run with a bad date hanging around so its appropriate to crash. Maybe we should guarantee that more explicitky

Pinpickle: Yeah should we crash at this stage instead of returning a time-bomb?

Sources/HaCWebsiteLib/ViewModels/LandingPage/LandingFeatures/GitWorkshop.swift
```diff
@@ -0,0 +1,19 @@
+import HaCTML
+import Foundation
+
+extension LandingFeatures {
+  static var gitWorkshop: LandingFeature {
+    return EventFeature(
+      startDate: Date.from(year: 2017, month: 11, day: 1, hour: 19, minute: 00),
+      endDate: Date.from(year: 2017, month: 11, day: 1, hour: 20, minute: 30),
+      eventLink: "https://www.facebook.com/events/1906244696304681",
```
moosichu: Do we possibly want two links? Like a pre-event link (for the Facebook event), and the on-event link (for the Workshop page?)

Apart from that, this all looks good, and is already a massive improvement over what we have!

Sources/HaCWebsiteLib/Utils/DateUtils.swift
jaredkhan: Yeah I think people are too inconsistent with date strings. Will probably end up with months and days mixed up at some point. This constructor seems readable enough to me

Sources/HaCWebsiteLib/ViewModels/LandingPage/LandingFeatures/GitWorkshop.swift
jaredkhan: See `liveLink`

Tests/HaCWebsiteLibTests/DateUtilsTests.swift
jaredkhan: Maybe this is a bridge we should cross when we come to it. Unless there's an obvious alternative way to test this?

Pinpickle: I can't think of an obvious alternative - I like the idea of waiting for a bridge.

jaredkhan:
@Pinpickle Have avoided the use of implicitly unwrapped optionals and explicitly used fatalError instead. The overhead required to test this is wacky so I've unfortunately omitted that test but I don't think this should be a huge concern since with luck we'll plan most of our events at times that exist.

PR #120, opened 2017-11-09
Add support for events in database

### LINKED ISSUE #23: Need a database representation of 'Events'

CharlieCrisp:
This PR changes our model of `Events` such that they are now `class`s not `struct`s. This PR also merges the granular `Event` classes into one large, multipurpose `GeneralEvent` so as to reduce complexity in the database. 

This PR no longer dumps the error when we try to initialise a pre-existing table (although an error is still bubbled up directly from the database)

This PR also extends the API so that the server can receive `HTTP` `POST`s, authenticate them against a password in the `.env` file and then dynamically add events to the database.

The implementation in this PR will check that all required fields have been provided in the `POST` request and will provide 'helpful' messages if they have not.

Fixes #23


moosichu:
Gonna look at this in depth tomorrow, have fewer time pressures

Sources/HaCWebsiteLib/Controllers/EventApiController.swift
```diff
@@ -0,0 +1,103 @@
+import Kitura
+import KituraNet
+import SwiftyJSON
+import Foundation
+import DotEnv
+
+struct EventApiController {
```
Pinpickle: I like that we're coming up with this whole API thing to start off with. It's going to make website maintenance much easier and less error prone.

Sources/HaCWebsiteLib/Controllers/EventApiController.swift
```diff
@@ -0,0 +1,103 @@
...
+  static var handler: RouterHandler = { request, response, next in
+    response.headers["Content-Type"] = "text/plain; charset=utf-8"
+    guard let parsedBody = request.body else {
```
Pinpickle: You've done an awesome job of figuring out all the boilerplate required. Hopefully we can abstract this all away in the future!

Sources/HaCWebsiteLib/Controllers/EventApiController.swift
```diff
@@ -0,0 +1,103 @@
...
+
+  static func isCorrectPassword(json: JSON) -> Bool {
+    let password = json["password"].stringValue
+    guard let realPassword = DotEnv.get("API_PASSWORD") else {
+      return false
+    }
+    return password == realPassword
+  }
+
+  static func parseEvent(json : JSON) throws -> GeneralEvent {
```
Pinpickle: Would we be able to save some work by making events conform to `Codeable`?

Relevant docs: https://developer.apple.com/documentation/foundation/archives_and_serialization/encoding_and_decoding_custom_types

Sources/HaCWebsiteLib/Controllers/EventApiController.swift
```diff
@@ -0,0 +1,103 @@
...
+
+  static func verifyAndSaveEvent(json: JSON) throws {
+    if !isCorrectPassword(json: json) {
+      throw EventParsingError.incorrectPassword
+    }
+    let event = try parseEvent(json: json)
+    try event.save()
+  }
+
+  static func isCorrectPassword(json: JSON) -> Bool {
```
Pinpickle: Not a blocker but conventionally information like passwords go in request headers, not in the body.

Would it make more sense to move this into a middleware so API endpoints automatically get this functionality?

CharlieCrisp: Looks like we can just use [this](https://github.com/IBM-Swift/Kitura-CredentialsHTTP) so I'll look into that

Sources/HaCWebsiteLib/Controllers/EventApiController.swift
```diff
@@ -0,0 +1,97 @@
...
+    let event = try parseEvent(json: json)
+    try event.save()
+  }
+
+  static func parseEvent(json : [String : Any]) throws -> GeneralEvent {
+    let dateFormatter = DateFormatter()
+    dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.zzz"
+
+    guard let title = json["title"] as? String,
+    let startDateString = json["startDate"] as? String,
```
Pinpickle: @CharlieCrisp I know `Encodeable` is giving you grief atm but would be worthwhile exploring later. Could you open an issue to further explore doing this all as encodeable?

Sources/HaCWebsiteLib/Utils/customEvent.json
```diff
@@ -0,0 +1,19 @@
+{
```
Pinpickle: What's your opinion on putting this and the test script in the repo?

I think this example is useful documentation, so maybe we do want to do it.

However, this will encourage people using the script to actually make changes, which is fine. Problem is, that they'll edit this file and I can see it accidentally getting committed and it's just another thing we have to spot in code reviews.

I think a markdown file with an example curl/postman/whatever script would prevent this. The user would have to copy and paste, then make their changes. It also allows us to document our API better.

Sources/HaCWebsiteLib/Utils/DatabaseUtils.swift
```diff
@@ -28,34 +28,36 @@ extension URL {
...
-  do {
-    let testPet = Pet(name: "Fluffs", age: 1)
-    try testPet.save()
-  } catch {
-    print("Failed to add record")
-    dump(error)
-  }
-}
\ No newline at end of file
+/*
```
Pinpickle: Is there a better place we can store this documentation? Such as on the docblock of the `GeneralEvent` class itself?

cc @jaredkhan for opinions on Swift and documentation and the like

Sources/HaCWebsiteLib/Utils/DatabaseUtils.swift
```diff
@@ -28,34 +28,36 @@ extension URL {
...
+}
 
-    let driver = try Driver(
-      masterHostname: databaseURLComponents.host,
-      readReplicaHostnames: [],
-      user: databaseURLComponents.user,
-      password: databaseURLComponents.password,
-      database: databaseURLComponents.database
-    )
+func prepareDatabase() {
```
Pinpickle: We're trying to avoid public/internal top level functions (correct @jaredkhan?) as it's not really idiomatic Swift (fine in the scope of a single file).

Could we namespace this in an enum?

Sources/HaCWebsiteLib/Utils/DatabaseUtils.swift
```diff
@@ -28,34 +28,36 @@ extension URL {
...
 }
 
-func testDatabase() {
-  do {
-    guard let databaseURLString = DotEnv.get("DATABASE_URL"),
-      let databaseURL = URL(string: databaseURLString),
-      let databaseURLComponents = databaseURL.databaseURLComponents else {
-      fatalError("Couldn't find valid DATABASE_URL environment variable")
-    }
+func getDatabaseDriver() throws -> PostgreSQLDriver.Driver {
```
Pinpickle: It appears that this is only used in this file. To this effect, could we make it private?

Sources/HaCWebsiteLib/Events/GeneralEvent.swift
```diff
@@ -1,44 +1,125 @@
...
+    get {
+      return self.hypePeriod.contains(Date())
+    }   
+  }
+  var postCardRepresentation : PostCard {
+    return PostCard(
+        title: self.title,
+        category: .general,
+        description: self.tagLine,
+        backgroundColor: self.color, //TODO
```
Pinpickle: What's this TODO referring to?

Sources/HaCWebsiteLib/Controllers/EventApiController.swift
CharlieCrisp: Issue has been opened :)

Sources/HaCWebsiteLib/Events/GeneralEvent.swift
CharlieCrisp: @Pinpickle this is from before when I think we were floating the idea of not using strings for colors. Maybe this deserves to be an Issue instead?

Sources/HaCWebsiteLib/Utils/DatabaseUtils.swift
CharlieCrisp: Ah yes - to be honest I'd forgotten this was in there and it probably needs to be taken out. It's very easy to figure out with a simple google so don't really think it belongs in our documentation

Sources/HaCWebsiteLib/Utils/customEvent.json
CharlieCrisp: I think that's a good idea :)

CharlieCrisp: @Pinpickle do you think this markdown would belong in a different directory?

Pinpickle: Yeah maybe a `docs/` folder?

Sources/HaCWebsiteLib/Events/GeneralEvent.swift
Pinpickle: Yeah, sounds like a plan

Sources/HaCWebsiteLib/Utils/DatabaseUtils.swift
jaredkhan: Yeah whacking it in a DatabaseUtils enum could be nice

Docs/Api/addEventSample.md
```diff
@@ -0,0 +1,28 @@
+### Example curl command to add an event from a JSON file in the same directory
```
Pinpickle: Building some docs, awesome!

These probably need fleshing out in the future but this will do nicely for now.

PR #126, opened 2017-11-10
Add pizzazz to About headers

jaredkhan:
Adds a little glam to the landing page headlines

Previously
screen shot 2017-11-10 at 22 37 13

Now
screen shot 2017-11-10 at 22 37 25


CharlieCrisp:
I like it! Is this going to be affected by the changes you were sketching out the other day @jaredkhan?

jaredkhan:
@CharlieCrisp Maybe but I think these headers might stick around. Either way, I think this is nicer in the meantime 😊

moosichu:
I need to look at in the context of the actual page. I really like the font, but this also lowers the information density of the page. (The headers taking up more space now). However, I do like it a lot. @Pinpickle Any comments?

jaredkhan:
@moosichu You're right but I don't think information density was ever particularly a goal

Pinpickle:
@moosichu we say information density like more is better but I don't think that's necessarily true.

However, I think in the context of smaller screens (like this screenshot), the headings are pretty overpowering. @jaredkhan what do you think of a media query to make them a touch smaller on smaller screens?

Something else that might help here is giving the headings a little more room to breath, ie some extra margin below them.

jaredkhan:
Yeah we could make them smaller on smaller screens. I think perhaps the column breakpoint is too small as well.

jaredkhan:
@Pinpickle @moosichu How's that?

Pinpickle:
@jaredkhan can we get some screenies?

jaredkhan:
I'll do you one better. Have 10 screenies per second
http://recordit.co/TXgxbGwdYT

PR #128, opened 2017-11-11
Landing copy

jaredkhan:
Uses the Markdown node introduced in #127 to simplify the Landing copy


moosichu:
Also adds a link to the demonstrators group?

jaredkhan:
@moosichu Yep, that link was supposed to be there before. Fixes a couple other typos too.

moosichu:
Also, now you can see why I love the new swift 4 multiline strings

CharlieCrisp:
This is really nice!

PR #131, opened 2017-11-11
Colocate render methods with their types

### LINKED ISSUE #129: Code for `Node` conformance should be kept with the type itself
## What the enhancement is
Just a code reshuffling, to avoid devs hunting for where the render function is for various Nodes (they're currently all together in render.swift)

jaredkhan:
Fixes #129


Sources/HaCTML/HTMLElement.swift
```diff
@@ -54,3 +56,24 @@ public struct HTMLElement {
     return clone(attributes: self.attributes.merge(attributes: attributes))
   }
 }
+
+extension HTMLElement: Node {
```
Pinpickle: I like that we're still keeping these as extension methods

Sources/HaCTML/AttributeMap.swift
```diff
@@ -37,7 +37,26 @@ struct AttributeMap: ExpressibleByArrayLiteral {
...
+    }
+
+    return renderedValues.flatMap(AttributeMap.renderAppliedAttribute).joined(separator: " ")
+  }
+
   var isEmpty: Bool {
     return self.attributeMap.isEmpty
   }
+
+  static func renderAppliedAttribute(_ attribute: AppliedAttribute) -> String? {
```
Pinpickle: I feel like adding these to the struct here clutter it, but there isn't much else to do and if it makes the codebase easier to parse then I'm for it.

jaredkhan: Agreed this is not an ideal place for this but a better one didn't come to mind. I am keen to avoid top-level functions  in packages where possible, I think mostly from a hygiene point of view.
Overall, now that we're on Swift 4, perhaps we could take another look at the layout of the Attribute stuff and tidy a few things up, I still find it tricky to navigate all these different Attribute bits and pieces

jaredkhan:
@Pinpickle To confirm, you happy with this going in?

Pinpickle:
@jaredkhan yeeeap

moosichu:
@jaredkhan @Pinpickle My understanding was that you are only meant to 'approve' a pull request under those circumstances. :wink: :stuck_out_tongue_winking_eye:

jaredkhan:
@moosichu True, we're a little too relaxed about this maybe. But also there's a difference between being happy for it to be merged, and being happy for it to be merged _right now_

PR #134, opened 2017-11-13
Add 'Today' label for events on the same day

### LINKED ISSUE #122: 'Today' Label on features that are relevant today
## What the enhancement is
Our EventFeatures currently have a 'On now' label when they are currently happening
it would be nice if they also had a 'Today' tag when they are happening the same day

## Why it could be helpful
People aren't good at understanding dates at a glance

martinhartt:
Fixes #122

Previously
screen shot 2017-11-14 at 14 03 19

Now
screen shot 2017-11-14 at 14 03 42


Sources/HaCWebsiteLib/Controllers/EventBetaController.swift
```diff
@@ -0,0 +1,23 @@
...
+import Foundation
+import HaCTML
+
+func getDate() -> Date {
+  let formatter = DateFormatter()
+  formatter.dateFormat = "yyyy/MM/dd HH:mm"
+  return formatter.date(from: "2017/12/13 14:31") ?? Date()
+}
+
+struct EventBetaController {
```
jaredkhan: Did this end up being useful for testing or just an extra hassle?

Sources/HaCWebsiteLib/ViewModels/LandingPage/EventFeature.swift
```diff
@@ -31,6 +31,10 @@ struct EventFeature: LandingFeature {
       return El.Div[Attr.className => "EventFeature__date EventFeature__date--live"].containing(
         "On now"
       )
+    } else if isToday {
+      return El.Div[Attr.className => "EventFeature__date EventFeature__date--live"].containing(
```
jaredkhan: Now that the `EventFeature__date--live` class applies to more than just when the event is 'live' (i.e. happening right now), perhaps we should rename it (e.g. to `EventFeature__date--highlight`)

Sources/HaCWebsiteLib/ViewModels/LandingPage/EventFeature.swift
```diff
@@ -56,6 +60,11 @@ struct EventFeature: LandingFeature {
     return paddedEventPeriod.contains(currentDate)
   }
 
+  // Whether the event is on the same day
+  var isToday: Bool {
+    return NSCalendar.current.isDateInToday(eventPeriod.start)
```
jaredkhan: Neat! 😊

Sources/HaCWebsiteLib/ViewModels/LandingPage/EventFeature.swift
```diff
@@ -56,6 +60,11 @@ struct EventFeature: LandingFeature {
     return paddedEventPeriod.contains(currentDate)
   }
 
+  // Whether the event is on the same day
```
jaredkhan: To make this into a [Markup Formatted](https://developer.apple.com/library/content/documentation/Xcode/Reference/xcode_markup_formatting_ref/SingleLineComment.html#//apple_ref/doc/uid/TP40016497-CH102-SW1) documentation note, add another slash (i.e. `///`). This is so that tools like Xcode know that it's a user-facing documentation comment rather than a comment for people looking at the source file.

jaredkhan:
Another thing, for changes that affect the frontend, please add screenshots 😊

martinhartt:
Thanks @jaredkhan! 

With regards to the codebase, it took me a while to understand what each subfolder was for (HaCTML vs HACWebsite vs HACWebsiteLib and Controllers vs Events vs ViewModels vs Views). Most of them were pretty self-explanatory, but for some I had to skim through the code to understand their purpose. Maybe some high-level documentation of the code structure would be helpful?

Sources/HaCWebsiteLib/Controllers/EventBetaController.swift
martinhartt: It was useful for testing whether the "Today" applied for specific dates and times. Should I remove this from the PR before it's merged?

Sources/HaCWebsiteLib/ViewModels/LandingPage/EventFeature.swift
martinhartt: Sure that makes sense. So should I add the highlight class only for the "Today" label?

Sources/HaCWebsiteLib/ViewModels/LandingPage/EventFeature.swift
martinhartt: Of course I totally forgot about this 😄

Sources/HaCWebsiteLib/Controllers/EventBetaController.swift
jaredkhan: Perhaps let's take this out for production yeah. It's worthwhile playing with these sandbox things though, I think they could be a lot more useful in the future.

Sources/HaCWebsiteLib/ViewModels/LandingPage/EventFeature.swift
jaredkhan: Since both "Today" and "On now" should be highlighted, let's use the highlight class for both and change the CSS accordingly

jaredkhan:
Thanks @martinhartt that's good feedback

Sources/HaCWebsiteLib/ViewModels/LandingPage/EventFeature.swift
Pinpickle: What timezone is this in?

martinhartt: @Pinpickle This is a good point, as right now it depends on which server this is running on. I updated it so that the calendar has the GMT timezone.

moosichu:
@martinhartt Feel free to merge this whenever! 👍

PR #137, opened 2017-11-15
Bump Kitura to v2.0.0

CharlieCrisp:
This PR will bump Kitura up to v2.0.0.
The reason for this is that the old version of Kitura causes a dependency resolution conflict when trying to add the Credentials dependency.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -29,21 +29,21 @@
         }
       },
       {
-        "package": "CPostgreSQL",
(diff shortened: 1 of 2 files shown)


moosichu:
@CharlieCrisp Why does it also add SwiftyJSON?

CharlieCrisp:
@moosichu SwiftyJSON used to be a dependency of Kitura. They have removed it as a dependency for `v2.0.0` but it's still in use in our codebase.

PR #140, opened 2017-11-18
Add webpack for cache busting

### LINKED ISSUE #51: Optimise production build with cache-busting

Pinpickle:
Fixes #51 

This adds Webpack to our build pipeline, which gives us cache busting and quite a lot of expressivity for any future static asset building we may want to do.

Assets are now resolved (in Swift) with `Asset.publicPath("/styles/main.css")`. This may seem like a pain, but it also makes it much easier for us to use a CDN if we ever want to.

In production, file paths are mangled like so:

```
<head><meta charset="UTF-8"><meta content="width=device-width, initial-scale=1" name="viewport"><link href="/static/images/favicon.a60f25a665cd78b2a8a2015bfc9a0103.png" type="favicon/png" rel="icon"><title>Hackers at Cambridge</title><link href="/static/styles/main.b1cb6b4ce736618531c8.css" type="text/css" rel="stylesheet"></head>
```

This doesn't add minification as it's not a priority, I'll split up the existing issue


Sources/HaCWebsiteLib/Controllers/EventStreamController.swift
```diff
@@ -18,14 +18,14 @@ struct LandingUpdateFeedController {
...
-      imageURL: "/static/images/functions_frame.png"
+      imageURL: Assets.publicPath("/images/functions_frame.png")
     ),
     PostCard(
       title: "TCP Throughput",
       category: .video,
       description: "Learn all the things.",
       backgroundColor: "green",
-      imageURL: "/static/images/workshop.jpg"
+      imageURL: Assets.publicPath("/images/workshop.jpg")
```
moosichu: Wouldn't it make sense to just give the `imageUrl `as a string to `PostCard`, and have it wrap it with `Assets.publicPath`, makes code a bit more dry no?
In face, couldn't we implement this as an extension to `String`, and have imageUrl treat it as that type?

Pinpickle: What happens if we want to link to an external image that we don't host?

`Assetable` is worth thinking about though, but that adds a lot of burden to the implementers of these kinds of components (and we tend to make a lot of them)

moosichu: @Pinpickle If it's an external image, won't the url be of the format `http://blah.blah`, it shouldn't be that difficult to treat those differently right?

Pinpickle: Yeah that is something you could reasonably check for. However, pattern matching on strings kind of goes against all of this nice type checking we're trying to do. I can't imagine a case at the moment, but I don't see it as unthinkable that we want a URL with a leading slash that isn't an asset. Maybe, for example, we have an endpoint that generates them dynamically.

This issue of burden of implementation is a big one. For this idiom to be useful, every component that has an input of an asset of _any_ kind will need to have this layer. This isn't even possible as our lowest level elements from `HaCTML` can't have it. I can imagine this leading to confusion. This is more to think about as implementors, and more to check for as reviewers

It also tightly couples our UI components to our build and asset system. If we ever wanted to make a library of reusable components, we'd be in trouble.

moosichu:
```swift
extension String: Assetable {
  var assetPath: String {
  if isRelativeString(self) {
    return Asset.publicPath(self)
  } else {
    return self
  }
}
```

moosichu: How does this add burden?

moosichu: Just saw your new comment, reading now!

Sources/HaCWebsiteLib/routes.swift
```diff
@@ -17,7 +17,7 @@ func getWebsiteRouter() -> Router {
     "/git": "https://github.com/hackersatcambridge/git-workshop-2017"
   ]))
 
-  router.all("/static", middleware: StaticFileServer(path: "./static/dist"))
+  Assets.addTo(router: router)
```
jaredkhan: Don't love that we're obscuring the fact that we're adding middleware here. Is there any benefit to having an `addTo` function? Will this ever be used more than once in a given codebase?

Pinpickle: This was so we can abstract the fact that there is this `/static` route, and that `Assets` can own it completely. Developers don't need to know that this thing exists

Pinpickle: Why is obscuring adding to middleware necessarily a bad thing?

gulpfile.js
```diff
@@ -1,41 +1,83 @@
...
-      console.log(err.message)
-      console.log('\u0007')
-      this.emit('end')
-    })
-    .pipe(autoprefixer())
-    .pipe(gulp.dest('static/dist/styles'))
+function buildStatic () {
+  // To prevent lots of files and add ease to live reloading,
+  // We don't want to hash filenames in dev mode
+  const dotHash = isProduction ? '.[hash]' : ''
```
jaredkhan: What's this square bracket syntax?

Sources/HaCWebsiteLib/routes.swift
moosichu: I agree with @jaredkhan, it does make it seem a bit 'magic'.

However, looking at it again through the idiomatic Swift lens, it does make more sense. I don't think it's too bad.

Sources/HaCWebsiteLib/Controllers/EventStreamController.swift
Pinpickle: I think remembering to do this for anything that can be an asset is just another place where we can make mistakes.

But my bigger worry is that it's impossible to do this in all parts of the codebase. HaCTML elements were one example. Another is where we're not necessarily rendering to HTML (maybe we're rendering JSON or making an API call). Users will need to remember to wrap their assets in `publicPath` sometimes but not in others.

Furthermore, I still believe there could be cases where we'd want to have a leading slash and not have the asset resolver step in.

Sources/HaCWebsiteLib/Assets.swift
```diff
@@ -0,0 +1,59 @@
...
+  }
+}
+
+private struct AssetResolver {
+  let manifest: JSON
+
+  init?(manifestPath: String) {
+    if let manifestFile = try? String(contentsOfFile: manifestPath, encoding: .utf8),
+       let manifestData = manifestFile.data(using: .utf8, allowLossyConversion: false) {
+      manifest = JSON(data: manifestData)
```
Pinpickle: Testing a comment

Sources/HaCWebsiteLib/Controllers/EventStreamController.swift
moosichu: @Pinpickle that's why there's a type system right? imageURL is specifically of type Assetable, because it is an asset. That's the documentation saying that you don't need to wrap it. It can also check if strings are valid URLs and complain if they aren't.

Not having it in HaCTML is fine, because that's basically everything in the rawest form.

And for the cases where you don't want the asset resolver to step in with a leading slash, you wouldn't have that variable be `Assetable` anyway right?

Pinpickle: The type system is kind of deceptive here, as `String`s are still `Assetable`, so we're not really getting any more type safety.

The down-level component can't know whether the consumer wants to resolve on a leading slash or not so it can't make that decision. We could add another type conforming to `Assetable` that has this behaviour but that is starting to seem like a lot of work.

What's your opinion on the tight coupling issue?

moosichu: I just saw that.  It is a good point. But I don't think makes it too hard to change to something reusable in the future. I believe what makes the codebase easier for us to use in the present is more important, but these tradeoffs are worth considering.

What if we created a new type `Asset` that can be initialised with strings? And why isn't the field `imageUrl` and `String?` type instead of a `URL?` type. The latter is more idiomatic no?

Sources/HaCWebsiteLib/Assets.swift
```diff
@@ -0,0 +1,59 @@
+import Kitura
+import SwiftyJSON
+import Foundation
+
+enum Assets {
```
jaredkhan: This file could use some docs

Sources/HaCWebsiteLib/routes.swift
jaredkhan: When adding things to the router, order matters. If we for (a contrived) example had a workshop on static analysis and tried to short link it to `/static`, we might skim the routes file to find out why this isn't working. Hiding this in Assets would make figuring this out harder.

gulpfile.js
Pinpickle: That's a webpack-ism: https://webpack.js.org/configuration/output/#output-filename

Sources/HaCWebsiteLib/Config.swift
```diff
@@ -24,4 +24,9 @@ public struct Config {
   public static func checkEnvVars() {
     checkEnvVarsExist(for: "DATA_DIR")
   }
+
+  public static var isProduction: Bool {
```
moosichu: We can use this now to silence compiler output on production now can't we?

moosichu: Should I create an issue for this?

Sources/HaCWebsiteLib/Controllers/EventStreamController.swift
jaredkhan: I agree that components shouldn't care about how the URL was created or whether it's local or remote, it should just accept a URL and not have to process it at all because this logic isn't specific to the component. 

On `String` vs `URL`, the reasoning here is a tradeoff between safety and productivity.
`URL(string: String)` would give us an _optional_ URL. It is nil if it's an invalid URL (i.e. doesn't conform to the grammar of a URL). In practice this means almost anything works (e.g. `hacttp://donu.ts`) works here. We'd be more interested in protecting against mistyped URLs that lead to nowhere (which this doesn't solve). 

Dealing with the optional from URL would practically mean force unwrapping (crashing at runtime if we mistype a URL). This might not be that useful.

moosichu: @jaredkhan So leave it as it is? :)

jaredkhan: IMO, yeah. 

We could revisit the URL type safety issue later.
For your interest: https://talk.objc.io/episodes/S01E71-type-safe-file-paths-with-phantom-types

Sources/HaCWebsiteLib/routes.swift
Pinpickle: @jaredkhan ideally we won't have `/static` hardcoded in even more places in our codebase. We could make the function name more informative?

`Assets.addFileServingMiddleware(to: router)`

Sources/HaCWebsiteLib/Config.swift
Pinpickle: @moosichu yeah go ahead!

Sources/HaCWebsiteLib/routes.swift
moosichu: @Pinpickle I would be happy with that
I was trying to think of funciont names to help with this, but came up blank. That looks good.

jaredkhan: @pinpickle my point was that middleware can basically hijack whatever is fed to it. I think hiding the fact that `/static` is in use for static assets is confusing

moosichu: @jaredkhan What if we explicitly passed `/static` to the `Assets` enum? What if it was a struct that was explicitly instantiated once, instead of an enum with global stuff going on?

jaredkhan: I think we should just pull this function outside of Assets. I disagree with your point @Pinpickle that 'Developers don't need to know that this thing exists'. It affects the routing behaviour on a path that they _might_ use so they _do_ need to know it exists.

Pinpickle: Okay I think I know how to go ahead and make this work for everyone, watch this space

Pinpickle:
@jaredkhan I've refactored the middleware stuff. Care to have another look?

Sources/HaCWebsiteLib/Assets.swift
```diff
@@ -0,0 +1,92 @@
...
+      return resolver.publicPath(assetPath)
+    }
+    
+    fatalError("Could not create AssetResolver")
+  }
+
+  /**
+    Initialize with configuration. This must be done exactly once before using the public API
+   */
+  public static func initialize(config: AssetsConfig) {
```
jaredkhan: Why not make Assets a struct with an initialiser?

Pinpickle: 2 reasons:

1. I feel like having a struct with both a static and instance API is confusing. Especially in this case when most of the state is static
2. This way the `urlBase` is kept private - only those configuring the assets can access it

There are ways to get around both of these issues while making it a struct, but this seemed pretty natural.

jaredkhan: Seems like we're jumping through a lot of hoops with optionals and force unwrapping here for no gain. Why not today this up by moving `resolver`, `config`, and `publicPath` to the instance level. This will also mean we can give Assets a failable initialiser that fails if AssetResolver initialiser fails so we don't have to hold on to an optional resolver either.

Pinpickle: `resolver` has to be kept separate from `config` as in development is re-created for every request. This is so the manifest can be re-read when the files we are loading changes.

This isn't necessary in production as the file is guaranteed to be the same for the lifetime of an instance, so it's disabled there to stop IO from clogging up our requests.

Pinpickle: Having an instance version of `Assets` won't stop us from force-unwrapping optionals won't it? It still has to be configured globally by some external thing

jaredkhan: I see what you're saying about a 'static API'. The issue is that we're using Asset.publicPath(...) in our viewModels right? So if we wanted to use an instance instead we'd need to hold a global instance somewhere. Am thinking of refactoring the routing stuff soon anyway so maybe we can revisit this.

Sources/HaCWebsiteLib/Assets.swift
```diff
@@ -0,0 +1,92 @@
...
+  func publicPath(_ assetPath: String) -> String {
+    if let publicPath = manifest[omitLeadingSlash(assetPath)].string {
+      return publicPath
+    }
+
+    fatalError("Asset not found: \(assetPath)")
+  }
+}
+
+private func omitLeadingSlash(_ string: String) -> String {
```
jaredkhan: Perhaps it'd make life simpler if we just disallowed leading slashes

Sources/HaCWebsiteLib/Assets.swift
Pinpickle: Yeah that's precisely the problem.

Sources/HaCWebsiteLib/Assets.swift
jaredkhan: Yeah but here they have the same meaning since we just ignore the leading slash. If I'm reading this right, `Asset.publicPath('images/test.png') == Asset.publicPath('/images/test.png')`

Pinpickle: Yeah it's a pretty thinly-veiled model. But people used to root relative URLs can come here and have it just work

Pinpickle: That also tightly couples us to the way webpack outputs manifests in a way. We're only doing that because it's convenient in this particular case

jaredkhan: Okay. As a style choice for this repo then, do we say that we should always include the learning slash?

Pinpickle: :+1:

PR #143, opened 2017-11-18
Update README with Windows hot-reload instructions

### LINKED ISSUE #77: Hot reloading doesn't work on Windows
Needs investigating

¯\\_(ツ)_/¯

moosichu:
Fixes #77, but once this is merged we should create a new issue for when [Linux containers on Windows](https://blog.docker.com/2017/09/preview-linux-containers-on-windows/) goes live, a this should enable hot-reloading out of the box. See docker/for-win#56 for more.


Pinpickle:
For us that don't understand the complex relationship between Docker, Linux and Windows: If Linux containers on Windows aren't a thing, what are you running your Docker containers on atm?

moosichu:
@Pinpickle Still Linux, but I think "Linux Containers on Windows" is a different thing? TBH I'm a bit confused by it. Maybe it's related to the environments the container is run in, a opposed to what the container contains??

PR #144, opened 2017-11-19
Add google analytics script 📈

### LINKED ISSUE #112: Add analytics
## What the enhancement is

Add analytics to our front end (such as Google Analytics) to track how many users our website has.

## Why it could be helpful

Allows us to measure the success of our website according to different metrics, so we know where to focus our efforts.

martinhartt:
Fixes #112 .

There are no visual changes, but there are changes to generated HTML.

Previously
screen shot 2017-11-19 at 09 56 20

Now
hacscreenshot

**IMPORTANT NOTE**: To enable proper functionality of analytics, add a new environment variable "GOOGLE_ANALYTICS_ID" containing the google analytics id in the form "UA-XXXXXXXXX-X"


Sources/HaCWebsiteLib/ViewModels/GAScript.swift
```diff
@@ -0,0 +1,22 @@
+import HaCTML
+import Foundation
+import DotEnv
+
+struct GAScript: Nodeable {
+  let googleAnalyticsId = DotEnv.get("GOOGLE_ANALYTICS_ID")!
```
Pinpickle: I love love love that we're putting this in an environment variable. Where this would be the most useful, I think, is in development environments where we do not want analytics to be recorded. However, this will crash when the variable is not present.

Would it be more useful to instead not add the script at all if there is no GA ID set? This way we can easily enable it for production and developers don't have to do anything.

martinhartt: This is a great point, you're absolutely right. I've updated the implementation to include this behaviour.

martinhartt:
@Pinpickle Sorry I messed up the merge.. Can you approve it once more please?

martinhartt:
@moosichu Thanks 😄

PR #150, opened 2017-11-21
Use http gzip compression for serving all resources

### LINKED ISSUE #18: Served content should be gzipped
https://github.com/IBM-Swift/Kitura-Compression

eliotlim:
Fixes #18 Served content should be gzipped.

Uses [Kitura-Compression](https://github.com/IBM-Swift/Kitura-Compression) to provide Compression middleware.
All resources will be served using the gzip standard by default.
At this stage, parameters used are defaults but can be configured at a later date.


Sources/HaCWebsiteLib/routes.swift
```diff
@@ -39,6 +40,9 @@ func getWebsiteRouter() -> Router {
 
   router.all("/", middleware: NotFoundMiddleware())
 
+  // Enable http gzip compression globally
+  router.all(middleware: Compression())
```
Pinpickle: Double checking that we want compression to be at the end of the middleware chain and not the beginning?

Might be worth adding a comment to that effect so people don't add things below it.

jaredkhan: @eliotlim Any follow up on this?

eliotlim: I have experimented with placing the compression middleware at various points in the middleware chain. Placing it before some of the middleware seems to cause page rendering to break, and placing it after most of the middleware (for instance, right after the `NotFoundMiddleware`) has no visible impact on the page load or content encoding in transit.

In the interest of best practice, I suppose placing it at the end of the middleware chain makes the most sense since compressing the content should occur transparently and be one of the last transformations prior to transmission. It also makes adding router middleware / routes in the future less likely to break things.

Perhaps as suggested, a comment could be added to explain compression being at the end of the middleware chain.

jaredkhan: Sounds good to me. Wanna get that comment in and resolve conflicts? then this will be good to go!

eliotlim:
I'm not sure why I was required to do a merge twice, but... I've added an explanatory comment!
Could you check if the comment is sufficiently explanatory?

Issue #40, opened 2017-08-28
Workshop launch tracker


Pinpickle:
This is tracking all of the issues we need to solve for us to get workshops into a useable state on the website.

- [x] #156 
- [x] #157 
- [x] #158 
- [x] #159 
- [x] #197 
- [x] #198
- [x] #199
- [x] #211

jaredkhan:
Broken down into:
#156 #157 #158 #159 #160 #161

Pinpickle:
@jaredkhan I edited this issue to make it a tracker for those ones

Pinpickle:
I'm moving #160 and #161 out of this as they are not blocking a launch of workshops

jaredkhan:
Workshops are launched! 🚀

PR #155, opened 2017-11-23
Abstract the thread-safe Git functionality

jaredkhan:
Prep for #40 

Abstracts the git functionality into a more reusable struct


moosichu:
@jaredkhan Why rename `GitUtils` to `GitUtil`?

Sources/HaCWebsiteLib/Utils/GitUtil.swift
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  init(remoteRepoURL: String, directoryName: String) {
+    serialQueue = DispatchQueue(label: "com.hac.website.\(directoryName)-pull-queue")
+    self.remoteRepoURL = remoteRepoURL
+    self.directoryName = directoryName
+  }
+
+  /// A safe version of `unsafeUpdate`
+  public func update(then callback: @escaping () throws -> Void = {}) rethrows {
```
moosichu: It this meant to be asynchronous? Because it doesn't appear to behave that way. What are the goals of this PR?

jaredkhan:
@moosichu Renamed GitUtils ⟶ GitUtil because it went from a static API to an instance API so I wanted to instantiate a _GitUtil_

Sources/HaCWebsiteLib/Utils/GitUtil.swift
jaredkhan: Good catch! This is intended to by synchronous and I think this makes sense since users of this API can always build asynchronous behaviour on top of that when they need it. I've fixed this up now.
The goals of this PR is to make it easy to pull many repos into disk and not worry about the dispatch stuff. The main focus was moving the dispatch queuing behaviour into the GitUtil rather than having 'WorkshopManager' deal with it directly. So now for a given repository, calling `update` a bunch of times will block until each `update` call has been able to execute sequentially.

jaredkhan:
Thanks @moosichu !

PR #163, opened 2017-11-24
Add a NewWorkshop model

### LINKED ISSUE #156: Model component for Workshops processing
## What the enhancement is
Make a new Workshop model that can represent all of the information in https://github.com/hackersatcambridge/workshop-example (e.g. description.md, setup_instructions.md etc.)

## Why it could be helpful
We'll need this for #40

jaredkhan:
Closes #156 

Adds the struct that models the new notion of a Workshop


jaredkhan:
@moosichu

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
```diff
@@ -0,0 +1,54 @@
...
+public struct NewWorkshop {
+  let title: String
+
+  /// People that have directly contributed content to this workshop
+  let contributors: [String]
+
+  /// Non-contributors we would like to thank for helping to inspire, develop or facilitate this workshop
+  let thanks: [String]
+
+  /// Links that attendees may which to read after the workhsop
```
CharlieCrisp: Spelling of 'wish' (sorry)

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
```diff
@@ -0,0 +1,54 @@
...
+  let furtherReadingLinks: [(String, URL)]
```
CharlieCrisp: Why are these not of type `Link`?

Sources/HaCWebsiteLib/Models/Link.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+/// Models a hyperlink
+struct Link {
```
CharlieCrisp: Is this actually being used anywhere in this PR?

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
```diff
@@ -0,0 +1,54 @@
...
+
+  /// A link to a video recording of this workshop
+  let recordingLink: URL?
```
CharlieCrisp: Minor but shouldn't we have any optional fields underneath all the mandatory ones?

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
```diff
@@ -0,0 +1,54 @@
...
+
+  /// A link to any slides for this workshop
+  let slidesLink: URL?
+
+  /// A set of keywords associated with the workshop
+  let tags: [String]
+
+  /// A string of a https://spdx.org/licenses/ license identifier
+  // TODO: Make a type safe version of this
+  let license: String
```
moosichu: @jaredkhan Is it ok if I get on this and push it to this PR?

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
```diff
@@ -0,0 +1,54 @@
...
+  let promoImageForeground: URL
+
+  /// The background for `promoImageForeground`
+  let promoImageBackground: Background
+
+  /// Prose describing the nature of the workshop
+  let description: Markdown
+
+  /// Prose describing who the workshop is aimed at
+  let whoItsFor: Markdown
```
moosichu: I can't seem to find this in the examples, and wont this end up possibly duplicating the prerquisites section? I don't see who this benefits enough for it to worth it really, as tags really handle most of the use cases here.

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
```diff
@@ -0,0 +1,54 @@
...
+  let furtherReadingLinks: [(String, URL)]
+
+  /// A link to a video recording of this workshop
+  let recordingLink: URL?
+
+  /// A link to any slides for this workshop
+  let slidesLink: URL?
+
+  /// A set of keywords associated with the workshop
+  let tags: [String]
```
moosichu: I was actually thinking about this a bit, we should probably make sure that when this system goes live, we have a `/tags` page that allows us to view all the tags created so far. That way when tagging a workshop as that list grows, people can look at the ones that exist so that they use the `canonical` versions.

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
```diff
@@ -0,0 +1,54 @@
...
+  let whoItsFor: Markdown
+
+  /// Prose describing what attendees should already know
+  let prerequisites: Markdown
+
+  /// Prose describing how attendees should prepare for the workshop
+  let setupInstructions: Markdown
+
+  /// The URL to the workshop repository
+  let repositoryURL: URL
```
moosichu: Do we want a full URL here? (All our workshops will be at `github.com/hackersatcambridge/workshop-name`. So maybe it should just be `workshopCanonicalName`.
Eg. For the love workshop that would be `love2d` when the URL is [`https://github.com/hackersatcambridge/workshop-love2d`](https://github.com/hackersatcambridge/workshop-love2d).

This also helps internally, as that will map to a local route for this of `https://hackersatcambridge.com/workshops/name` with a redirect to that from `https://hackersatcambridge.com/workshops/workshop-name` and `https://hackersatcambridge.com/w/name` .

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
jaredkhan: Let’s put it in a separate PR after this goes in since this blocks things

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
jaredkhan: Good point. I actually removed it from the example for this reason!

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
moosichu: The URL could be a derived property

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
jaredkhan: Don't think so necessarily. The default member-wise initialiser still makes us explicitly say `nil` for these so it doesn't make much difference. They're in this order because this is roughly the order they're in in the workshops repos.

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
jaredkhan: Sounds easy enough 👍

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
jaredkhan: Also, I reckon this can go in a separate module @moosichu so that we can move it to a separate repo eventually and other people can use it for their zesty projects 😊

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
moosichu: See issue #165, just created it.

jaredkhan:
@moosichu @CharlieCrisp Thanks for your reviews, think I've cleared up now

Sources/HaCWebsiteLib/Models/NewWorkshop.swift
moosichu: created #166 :)

Sources/HaCWebsiteLib/Models/Link.swift
```diff
@@ -0,0 +1,13 @@
...
+  let title: String
```
Pinpickle: If we're mapping this to HTML `a` elements, will `title` become its contents or the `title` attribute? [MDN on `title`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title)

Sources/HaCWebsiteLib/Models/Link.swift
jaredkhan: It is now! 😬

Sources/HaCWebsiteLib/Models/Link.swift
jaredkhan: I was thinking contents. Think title is a reasonable name for it. Would `text` be clearer? `linkText`?

moosichu: yeah, I think text is clearer

PR #175, opened 2017-11-30
Make Script loading portable

jaredkhan:
Previously were assuming we were running in `/root/hac-website`. Now we don't depend on that

### DIFF
--- a/Sources/HaCWebsiteLib/ViewModels/Script.swift
+++ b/Sources/HaCWebsiteLib/ViewModels/Script.swift
@@ -54,8 +54,10 @@ This class is used in order load front-end scripts from a file relative to the c
 struct Script : Nodeable {
   let file : String
   let definitions : [String: JavaScriptable]
-  let directory = "/root/hac-website/Sources/HaCWebsiteLib/ViewModels"
+
   var node: Node {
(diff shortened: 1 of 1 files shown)


Sources/HaCWebsiteLib/ViewModels/Script.swift
```diff
@@ -54,8 +54,10 @@ This class is used in order load front-end scripts from a file relative to the c
 struct Script : Nodeable {
   let file : String
   let definitions : [String: JavaScriptable]
-  let directory = "/root/hac-website/Sources/HaCWebsiteLib/ViewModels"
+
   var node: Node {
+    let websiteRoot = FileManager.default.currentDirectoryPath
+    let directory = websiteRoot + "/Sources/HaCWebsiteLib/ViewModels"
```
moosichu: Is there a reason why you do this instead of just:

`let directory = FileManager.default.currentDirectoryPath + "/Sources/HacWebsiteLib/ViewModels"`

jaredkhan: No reason other than to inform the reader that this `currentDirectoryPath` thing points to the root of the project, not the directory that this file is in.

PR #184, opened 2017-12-05
Source events from database

### LINKED ISSUE #130: Redefine EventServer.swift
We currently source our event stream from an `EventServer` struct. Internally, this class currently just serves hard-coded events. Once #120 is merged, we need to redefine this to source events from our database.

CharlieCrisp:
Fixes #130.

Our `EventServer` is responsible for providing all the upcoming (and past) events. Previously, it was hardcoded to provide some dummy events.

This PR fixes this by sourcing the events from a database. This is done whenever a request is made to the site. Specifically `/beta/landing-update-feed` at the moment.


Sources/HaCWebsiteLib/Events/EventServer.swift
```diff
@@ -2,22 +2,25 @@ import Foundation
...
+    events = notNilEvents
+  }
+
+  static func getAllEvents() -> [GeneralEvent] {
+    return events
+  }
+
+  static func getCurrentEvents() -> [GeneralEvent] {
+    return events.filter{ event in 
+      let currentDate = Date()
```
moosichu: Why are we doing this? Surely plan is to show all future events on the landing page. Also, surely doing this through a database search is faster that filtering through an in-memory array?

Sources/HaCWebsiteLib/Events/EventServer.swift
```diff
@@ -2,22 +2,25 @@ import Foundation
...
-        return GeneralEvent(title: workshop.title, time:  DateInterval(start: Date(), duration: duration),
-          tagLine: "There's Pizza", color: "purple", hypePeriod: DateInterval(start: Date(), duration: hypeDuration),
-          tags: ["tag1", "tag2"], description: Markdown("Thing"))
-    }.filter{ event in 
-      event.shouldShowAsUpdate
+  private(set) static var events: [GeneralEvent] = []
+
+  static func update() {
+    let newEvents = try? GeneralEvent.makeQuery().filter("title", .notEquals, nil).all()
+    guard let notNilEvents = newEvents else {
```
moosichu: surely:
```swift
guard let notNilEvents = newEvents {
   events = notNilEvents
}
```
is a cleaner way of doing this whole function right?

Sources/HaCWebsiteLib/Events/EventServer.swift
```diff
@@ -2,22 +2,25 @@ import Foundation
...
-    let hypeDuration : Double = 4000000
-    let duration : Double = 4000
-    let workshopEvents : [GeneralEvent] = WorkshopManager.workshops.map{
-      (workshop: Workshop) -> GeneralEvent in
-        return GeneralEvent(title: workshop.title, time:  DateInterval(start: Date(), duration: duration),
-          tagLine: "There's Pizza", color: "purple", hypePeriod: DateInterval(start: Date(), duration: hypeDuration),
-          tags: ["tag1", "tag2"], description: Markdown("Thing"))
-    }.filter{ event in 
-      event.shouldShowAsUpdate
+  private(set) static var events: [GeneralEvent] = []
```
moosichu: Why are we caching? Have you profiled the performance here? If we have a lot of events, wont this potentially be slower?

Sources/HaCWebsiteLib/Events/EventServer.swift
jaredkhan: Don't think so, but:
```
events = newEvents ?? []
```
should do the job!

Sources/HaCWebsiteLib/Events/EventServer.swift
CharlieCrisp: I was just really following the `WorkshopManager` format tbh. Not really sure what is quicker but I kinda assumed a database read would be slow. I may very well be wrong and should probably look into this

Sources/HaCWebsiteLib/Events/EventServer.swift
CharlieCrisp: I do think we want to leave in the behaviour of defaulting to an empty array if the query doesn't return anything. That would require setting `events` to `[]` before the guard which isn't really any different to what's already there.

Sources/HaCWebsiteLib/Events/EventServer.swift
jaredkhan: WorkshopManager was mostly doing this because it has to do a whole bunch of http requests and parsing to update rather than a relatively clean db query. This does feel like premature optimisation probs.

moosichu: Databases are (should) be designed for this kind of thing! I believe if you want to cache database stuff you really need to use something comprehensive like [memcached](https://en.wikipedia.org/wiki/Memcached).

Paging @Pinpickle for an opinion on this.

Sources/HaCWebsiteLib/Events/EventServer.swift
CharlieCrisp: Well, we may not wish to start advertising for all future events, just the ones which are currently being pushed. I think that this probably goes best alongside a `getCurrentAndFutureEvents` function though

Sources/HaCWebsiteLib/Controllers/LandingUpdateFeedController.swift
```diff
@@ -8,10 +8,17 @@ import SwiftyJSON
 
 struct LandingUpdateFeedController {
 
-  static let eventPostCards : [PostCard] = EventServer.getEvents().flatMap{ event in 
+  static var eventPostCards : [PostCard] = EventServer.getCurrentEvents().flatMap{ event in
```
jaredkhan: Again, caching might be unnecessary here:

```
static var eventPostCards: [PostCard] {
  EventServer.getCurrentEvents().flatMap{ event in 
    event.postCardRepresentation
  }
}
```
Would be a nice alternative and then ditch the `update` entirely. We can always optimise this later if need be but even then it might be cleaner to have some kind of subscriber thing rather than occasional manual polling.

Sources/HaCWebsiteLib/Events/EventServer.swift
moosichu: I mean, I thought the plan is to eventually show a general `event-stream` on the landing page, grabbing information from all our content: blog-posts, workshops, hackathons, videos, some shoutouts etc. Possibly going into the past and future.

CharlieCrisp: On the landing page we will want to keep the number of things we display to a minimum so this will only be current events. Sure we may have past videos/blog posts but that's a completely different matter of `updates` not `events.

Sources/HaCWebsiteLib/Events/EventServer.swift
```diff
@@ -2,22 +2,25 @@ import Foundation
...
+
+  static func update() {
+    let newEvents = try? GeneralEvent.makeQuery().filter("title", .notEquals, nil).all()
```
jaredkhan: When would a title be nil? In GeneralEvent, this is non-optional

CharlieCrisp: Never - it's just a way of getting all the events from the database as it seems with `Fluent` you have to specify some form of filter :(

moosichu: @CharlieCrisp Probably worth commenting then!

jaredkhan: `all()` is a method on QueryRepresentable, doesn't look like we need that filter clause at all!

CharlieCrisp:
@jaredkhan @moosichu I have addressed your comments and am now just sourcing events from the database upon each request

Sources/HaCWebsiteLib/Controllers/LandingPageController.swift
```diff
@@ -29,9 +25,11 @@ struct LandingPageController {
     )
   ]
 
-  static let updates = eventPostCards + videos
-
   static var handler: RouterHandler = { request, response, next in
+    let eventPostCards : [PostCard] = EventServer.getCurrentEvents().flatMap{ event in
```
jaredkhan: Hmm we are duplicating the post card getting behaviour here. We should perhaps have the LandingPageController instantiate a LandingPageUpdateFeedController and pass that down to the LandingPage view model. But that can happen in another PR if you wanna run ahead with this!

jaredkhan:
Is there more to be done on this @CharlieCrisp or is it good to go?

CharlieCrisp:
@jaredkhan we're good to go - I'll file an issue for the `LandingPageController`stuff

PR #194, opened 2017-12-11
Prefix every process output line

### LINKED ISSUE #193: Not all log messages from the web-server are properly labelled
## Screenshot of bug:

image
See underlined log messages for this weird behaviour.

## Explanation of bug:

See #116, it is something to do with Gulp.

`[kitura]` should probably be changed to `[web-server]` for our internal logging as well.

jaredkhan:
Fixes #193 

Now every line output by (e.g.) the server process will be labelled appropriately where previously some lines were not being labelled which was confusing


jaredkhan:
@moosichu

PR #195, opened 2018-01-13
Workshops parsing

### LINKED ISSUE #157: Data Processing component for Workshops processing
## What the enhancement is
Relies on #156 
Create a parser that can take a path to a [workshop repo directory](https://github.com/hackersatcambridge/workshop-example/tree/master/info) and parse that information into a Workshop.

jaredkhan:
Fixes #157

Pulls the [workshops index](https://github.com/hackersatcambridge/workshops-refactor/blob/master/workshops.yaml)
Pulls each of the referenced workshops
Parses each of the workshops in to NewWorkshop models

Adds a few tests for this parsing.

Notable omissions at the moment:
- Updates to workshops are not pulled in
- We're currently using the RawGit development server which we will need to ditch before we use this for actual workshops
- We're not currently transforming the relative URLs in workshop notes so that images actually work.


Sources/HaCWebsiteLib/NewWorkshopManager.swift
```diff
@@ -0,0 +1,91 @@
...
+    let updateDispatchGroup = DispatchGroup()
+
+    for (repoName, repoUtil) in workshopRepoUtils {
+      updateDispatchGroup.enter()
+      updateQueue.async {
+        do {
+          repoUtil.update()
+          let updatedWorkshop = try NewWorkshop(localPath: repoUtil.localRepoPath)
+          workshops[repoName] = updatedWorkshop
+        } catch {
```
Pinpickle: I like how one workshop failing doesn't stop the other workshops from being parsed. It would be good (but not a blocker, perhaps create an issue to track this?), if there were a way for us to be notified in a more significant way if something goes wrong. e.g. on Slack, email, or if we had a secret website dashboard that we could check.

Sources/HaCWebsiteLib/Models/NewWorkshop+init.swift
```diff
@@ -0,0 +1,285 @@
...
+  notesMarkdown: "/content/notes/notes.md",
+  examplesDirectory: "/examples",
+  presenterGuide: "/content/presenter_guide.md",
+  setupInstructions: "/info/setup_instructions.md"
+)
+
+let validImageExtensions = ["jpg", "svg", "png", "gif"]
+
+extension NewWorkshop {
+  enum Error: Swift.Error {
```
Pinpickle: Descriptive, fine-grained errors 😍 

Good thing Swift makes it so easy.

Sources/HaCWebsiteLib/Models/NewWorkshop+init.swift
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  /// Create a NewWorkshop from a local path
+  init(localPath: String) throws {
+    workshopId = try NewWorkshop.getWorkshopId(localPath: localPath)
+    let remoteRepoUrl = "https://github.com/hackersatcambridge/workshop-" + workshopId
+
+    // TODO: Save the presenterGuide.
+    // This is currently run but not saved so that it still runs checks on the file
+    _ = try NewWorkshop.getPresenterGuide(localPath: localPath)
```
Pinpickle: Why are we not saving this?

Sources/HaCWebsiteLib/Models/NewWorkshop+init.swift
jaredkhan: Part of the motivation for these is the issue you mentioned of notifications. The initialiser here serves as a validator. When I was plugging I previous examples to the test cases I got nice errors that told me what the first error was. 
Keen for that to be utilised either through the website notifying or by making it easy to add this as a PR status check on workshop repos 👍

Sources/HaCWebsiteLib/Models/NewWorkshop+init.swift
jaredkhan: Good question. The format of this guide hadn’t been discussed very much so I didn’t save it but it’s no extra work to do so. So I will!

PR #196, opened 2018-01-13
Style the individual workshop page

### LINKED ISSUE #159: Design and implement a page to view specific workshops
## What the enhancement is
A page to view the details of a specific workshop. It should show any useful details from the workshop model e.g. description, prerequisites, set-up instructions

At this point it needn't show upcoming dates, this is a separate issue (see #161)

jaredkhan:
Fixes #159

Notable missing:
- Good handling of code snippets
- Embedding of any content (slides or recording), might not be particularly convenient to do this anyway

Now
screen shot 2018-01-13 at 23 48 27
screen shot 2018-01-13 at 23 48 38
screen shot 2018-01-13 at 23 48 50

### DIFF
--- a/Sources/HaCTML/Fragment.swift
+++ b/Sources/HaCTML/Fragment.swift
@@ -28,12 +28,12 @@
 public struct Fragment: Node {
   let nodes: [Node]
 
-  public init(_ nodes: Nodeable...) {
+  public init(_ nodes: Nodeable?...) {
     self.init(nodes)
   }
 
-  public init(_ nodes: [Nodeable]) {
-    self.nodes = nodes.map{ $0.node }
+  public init(_ nodes: [Nodeable?]) {
+    self.nodes = nodes.flatMap{ $0?.node }
   }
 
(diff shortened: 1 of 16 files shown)


Pinpickle:
Why wouldn't it be convenient to embed recordings? I can understand why slides would be clunky.

Confirming: this isn't using actual data yet, is it?

jaredkhan:
@Pinpickle I’m thinking it’s probably going to be more convenient for users to open the recording in a new tab so they can scroll the content while listening to the recording

Pinpickle:
@jaredkhan that's a reasonable thought. In my experience people have been "either or" with videos vs written content. Either the video has worked for them or the written content has (or they're in a particular mood for on or the other).

Let's go with the least-effort path for now, though (a link)

jaredkhan:
@Pinpickle True but I don’t think we should necessarily encourage that because I think it’ll make people less likely to engage in the exercises. A discussion for another time

PR #200, opened 2018-01-17
Serve images with RawGit's production URLs

### LINKED ISSUE #198: Use RawGit production links for workshop images
For background, see here: http://rawgit.com for the difference between RawGit's dev and production URLs

Currently we're using RawGit development URLs for serving images in workshops (search rawgit.com in the codebase to see where). This needs to be changed before people can visit the workshops pages.

This will likely involve adding to GitUtil to find the latest commit hash of each of the repos and then using this in the URL.

Pinpickle:
Fixes #198 

This adds support for fetching the SHA of the current master for workshop repos, and using that to generate a production URL for RawGit. Note that I haven't tested the fetching of the SHA, but I have added tests to make sure the data is used appropriately.

I also refactored the workshop init function a bit, to save ourselves from passing lots of data around.

### DIFF
--- a/Sources/HaCWebsiteLib/Models/NewWorkshop+init.swift
+++ b/Sources/HaCWebsiteLib/Models/NewWorkshop+init.swift
@@ -17,7 +17,7 @@ private let validImageExtensions = ["jpg", "svg", "png", "gif"]
 
 extension NewWorkshop {
   /// Create a NewWorkshop from a local path
-  init(localPath: String, headCommitSha: String?) throws {
+  init(localPath: String, headCommitSha: String) throws {
     workshopId = try NewWorkshop.getWorkshopId(localPath: localPath)
     let metadata = try NewWorkshop.getMetadata(localPath: localPath)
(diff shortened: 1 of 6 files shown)


Sources/HaCWebsiteLib/Models/NewWorkshop+init.swift
```diff
@@ -13,151 +13,154 @@ let filePaths = (
...
+  case invalidPromoImageBackgroundFormat
+  case missingNotes
+  case badURLInNotes(String)
+  case emptyExamples
+  case invalidURL(String)
+}
+
+private struct WorkshopBuilder {
+  let localPath: String
+  let commitSha: String?
```
jaredkhan: When would this be nil?

Pinpickle: If we ever wanted to do some testing, didn't have a commit sha on hand or something. It's tenuous

Want me to remove it?

jaredkhan: Can give dud hashes for testing, reckon can make this non-optional

PR #201, opened 2018-01-17
Hook up NewWorkshops data to View

### LINKED ISSUE #197: Hook up workshops data
We're well on our way to having pages that display workshop information nicely and a workshops controller that pulls in the workshops from the workshops index repo. The next step is to hook up the data to the views. Work involves:
- List the workshops in a sensible order on the workshops index page.
- Add a router at /workshops/<workshop-id> that gets the relevant workshop data and displays it

Note that the workshops index page currently features events rather than acting primarily as an index of the workshop. This decision may need to be revisited to make this task make sense.

jaredkhan:
Fixes #197 

- Plugs in the IndividualWorkshopPage view to the given data
- Plugs in the WorkshopsIndexPage to available data by removing the events schedule and replaced with a workshops archive for now. The schedule should be reintroduced later.

screen shot 2018-01-17 at 15 02 23

### DIFF
--- a/Sources/HaCWebsiteLib/Views/Workshops/WorkshopsIndexPage/WorkshopsIndexArchive.swift
+++ b/Sources/HaCWebsiteLib/Views/Workshops/WorkshopsIndexPage/WorkshopsIndexArchive.swift
@@ -1,0 +1,21 @@
+import HaCTML
+
+/// Above-the-fold for the WorkshopsIndexPage
+struct WorkshopsIndexArchive: Nodeable {
+  let workshops: [NewWorkshop]
+
+  var node: Node {
+    return El.Div[Attr.className => "WorkshopsIndexPage__archive"].containing(
+      El.H2[Attr.className => "Text--sectionHeading"].containing("Browse Workshops"),
+      El.Ul[Attr.className => "WorkshopsIndexPage__archive__list"].containing(
+        workshops.map { workshop in 
+          El.Li.containing(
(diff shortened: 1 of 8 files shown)


Sources/HaCWebsiteLib/Views/Workshops/WorkshopsIndexPage/WorkshopsIndexArchive.swift
```diff
@@ -0,0 +1,21 @@
...
+struct WorkshopsIndexArchive: Nodeable {
+  let workshops: [NewWorkshop]
+
+  var node: Node {
+    return El.Div[Attr.className => "WorkshopsIndexPage__archive"].containing(
+      El.H2[Attr.className => "Text--sectionHeading"].containing("Browse Workshops"),
+      El.Ul[Attr.className => "WorkshopsIndexPage__archive__list"].containing(
+        workshops.map { workshop in 
+          El.Li.containing(
+            El.A[Attr.href => "/beta/workshops/\(workshop.workshopId)", Attr.className => "WorkshopsIndexPage__archive__list__item"].containing(
```
Pinpickle: We should probably stop hard-coding our URLs so if they change (as they will with `/beta/workshop`), we don't miss them.

Not a blocker, but something that is as easy as having one `WebsiteLink` struct with static functions could work (there are more elegant solutions).

jaredkhan: Was thinking about how best to do this. I figured it wasn't a priority for now

PR #206, opened 2018-01-20
Keep workshops up-to-date

### LINKED ISSUE #199: Keep workshops data up to date
## What the enhancement is
Relevant after #195 is merged
Currently workshops are only pulled in on start up and not updated
We need workshop data to be up-to-date at all times.
This will likely involve polling for changes to the data (via `.update()` on `NewWorkshopsManager`). An alternative is GitHub webhooks but this would be poor practice since we then rely on us having only one server (Heroku spins up and balances load over multiple instances if you let it)

## Why it could be helpful
We need to be able to hot fix workshops content

jaredkhan:
Fixes #199

- Polls for updates to workshops
- Provides a password-protected update trigger endpoint

### DIFF
--- a/Sources/HaCWebsiteLib/Controllers/WorkshopsController.swift
+++ b/Sources/HaCWebsiteLib/Controllers/WorkshopsController.swift
@@ -23,4 +23,11 @@ struct WorkshopsController {
     }
   }
 
+  static var workshopUpdateHandler: RouterHandler = { request, response, next in
+    NewWorkshopManager.update()
+    try response.send(
+      "Workshops updated!"
+    ).end()
+  }
+
 }
--- a/Sources/HaCWebsiteLib/NewWorkshopManager.swift
+++ b/Sources/HaCWebsiteLib/NewWorkshopManager.swift
@@ -17,6 +17,9 @@ public struct NewWorkshopManager {
     attributes: .concurrent
   )
 
+  private static var pollSource: DispatchSourceTimer! = nil
+  private static var currentlyUpdating = false
+
   /// The list of Workshops as derived from the workshop repos last time they were processed. Keys are workshop repo names
   public private(set) static var workshops: [String: NewWorkshop] = [:]
 
@@ -24,8 +27,30 @@ public struct NewWorkshopManager {
   private static var workshopRepoUtils: [String: GitUtil] = [:]
 
   public static func update() {
-    updateRepoList()
-    updateWorkshopsData()
+    print("Updating workshops...")
+    // Prevent multiple updates at once (might happen if an update takes longer than the poll interval)
+    if !currentlyUpdating {
+      currentlyUpdating = true
+      updateRepoList()
+      updateWorkshopsData()
+      currentlyUpdating = false
+    }
+    print("Workshops updated")
+  }
+
+  public static func startPoll() {
+    let pollQueue = DispatchQueue(label: "com.hac.website.workshops-poll-timer", attributes: .concurrent)
+    pollSource?.cancel()
+    pollSource = DispatchSource.makeTimerSource(queue: pollQueue)
+    pollSource?.setEventHandler {
+      update()
+    }
+    pollSource?.schedule(
+      deadline: DispatchTime.now(),
+      repeating: .seconds(600),
+      leeway: .seconds(2)
+    )
+    pollSource?.resume()
   }
 
   /// Update the list of workshop repositories synchronously. This does not update the workshops themselves
@@ -68,6 +93,7 @@ public struct NewWorkshopManager {
     })
   }
 
+  /// Update the in-memory copy of the workshops synchronously
   public static func updateWorkshopsData() {
     let updateDispatchGroup = DispatchGroup()
 
--- a/Sources/HaCWebsiteLib/NewWorkshopManager.swift
+++ b/Sources/HaCWebsiteLib/NewWorkshopManager.swift
@@ -42,15 +42,15 @@ public struct NewWorkshopManager {
     let pollQueue = DispatchQueue(label: "com.hac.website.workshops-poll-timer", attributes: .concurrent)
     pollSource?.cancel()
     pollSource = DispatchSource.makeTimerSource(queue: pollQueue)
-    pollSource?.setEventHandler {
+    pollSource.setEventHandler {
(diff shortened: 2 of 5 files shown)


Sources/HaCWebsiteLib/NewWorkshopManager.swift
```diff
@@ -17,15 +17,40 @@ public struct NewWorkshopManager {
...
+      currentlyUpdating = false
+    }
+    print("Workshops updated")
+  }
+
+  public static func startPoll() {
+    let pollQueue = DispatchQueue(label: "com.hac.website.workshops-poll-timer", attributes: .concurrent)
+    pollSource?.cancel()
+    pollSource = DispatchSource.makeTimerSource(queue: pollQueue)
+    pollSource?.setEventHandler {
```
Pinpickle: When might `pollSource` be `nil` here?

Sources/HaCWebsiteLib/NewWorkshopManager.swift
```diff
@@ -17,15 +17,40 @@ public struct NewWorkshopManager {
...
 
   /// All the workshop repos that we know about. Keys are workshop repo names
   private static var workshopRepoUtils: [String: GitUtil] = [:]
 
   public static func update() {
-    updateRepoList()
-    updateWorkshopsData()
+    print("Updating workshops...")
+    // Prevent multiple updates at once (might happen if an update takes longer than the poll interval)
+    if !currentlyUpdating {
```
Pinpickle: Are there any concurrency issues involved here?

Sources/HaCWebsiteLib/Controllers/WorkshopsController.swift
```diff
@@ -23,4 +23,11 @@ struct WorkshopsController {
     }
   }
 
+  static var workshopUpdateHandler: RouterHandler = { request, response, next in
+    NewWorkshopManager.update()
```
Pinpickle: Does this perform an update or queue one? If the former, note that Heroku will always time out if a request takes more than 30s (but updates shouldn't take that long?)

Sources/HaCWebsiteLib/NewWorkshopManager.swift
```diff
@@ -17,15 +17,40 @@ public struct NewWorkshopManager {
...
+      currentlyUpdating = true
+      updateRepoList()
+      updateWorkshopsData()
+      currentlyUpdating = false
+    }
+    print("Workshops updated")
+  }
+
+  public static func startPoll() {
+    let pollQueue = DispatchQueue(label: "com.hac.website.workshops-poll-timer", attributes: .concurrent)
```
Pinpickle: Not a blocker but might be nice to abstract this polling behaviour away. If only `RxSwift` worked 😢

Sources/HaCWebsiteLib/NewWorkshopManager.swift
jaredkhan: Never

Sources/HaCWebsiteLib/NewWorkshopManager.swift
jaredkhan: The only race is if a user triggers an update at the same time as an automatic update. This is mostly an optimisation trying to protect from queuing updates that we don't need rather than any real safety concern. This is unlikely to ever cause a problem but I can make this more robust and solve the polling abstraction thing below by creating a PollUtil that handles all this for us.

Sources/HaCWebsiteLib/NewWorkshopManager.swift
Pinpickle: Might we be able to replace `?` with a `!` (so the intent is clear) or make it not `nil` (e.g. with a `let newPollSource = ...`)

Sources/HaCWebsiteLib/NewWorkshopManager.swift
Pinpickle: We can solve it later. RxSwift's scheduling doesn't work (as far as I can tell), but we can use their observable operators for stuff like this.

Sources/HaCWebsiteLib/Controllers/WorkshopsController.swift
jaredkhan: This performs one. Will change this

Sources/HaCWebsiteLib/NewWorkshopManager.swift
jaredkhan: Is this going to add a lot of effort for new contributors?

Pinpickle: A good debate that I don't have answers for at the moment. On the surface I'm inclined to say other concurrency primitives are even more difficult to wrap your head around and the declarative nature of Rx and similar libraries make code easier to comprehend.

jaredkhan: Maybe we should run with something like this and then use RxSwift in the near future. Is there a way we could contribute a linux workaround for RxSwift scheduling?

Pinpickle: Yeah I definitely think we should just roll with this at the moment.

I think contributing something to fix the Linux stuff (either through documentation or actually fixing it) would be great

Sources/HaCWebsiteLib/Controllers/WorkshopsController.swift
jaredkhan: Actually this was by design so that the user can load this page, and then go reload the workshops pages to see their changes right away. Maybe should wait and see if this is an issue in practice? I honestly am not sure how long updates will practically take.

Pinpickle: Okay let's fly and see how this does.

Sources/HaCWebsiteLib/NewWorkshopManager.swift
jaredkhan: Okay cool. As I say, as far as I'm aware this is not a safety concern. Data updates are explicitly serialised. Index updates will have a race but that doesn't result in any crash

jaredkhan:
Anything blocking this?

PR #207, opened 2018-01-21
Add relative URL resolution to Markdown

### LINKED ISSUE #192: Markdown transformations for workshop notes
Related to issue #157 
## What the enhancement is
Workshop content now live in standalone repositories, e.g. https://github.com/hackersatcambridge/workshop-example

A principal piece of this content is the notes which accompany the workshop. One can embed images in these notes using relative URLs and storing the images in the images directory. See [the example notes](https://github.com/hackersatcambridge/workshop-example/tree/master/content/notes).

Currently, images embedded in workshop [notes] can't be displayed on the website. To make this possible we will want to copy images from the images directory to some place they can be served to the public (it's probably bad luck to serve them directly from GitHub) and then rewrite the markdown to point to the public copies of the images instead of using the relative URLs.

Roughly what this involves:

      // get all image URLs in notes.md
        // If relative URL
          // check that image exists
          // copy image over
          // rewrite the markdown
        // If absolute
          // Check that the URL is reachable, throw error if not
      // get all hyperlink URLs in notes.md
        // If relative
          // Throw an error, Notes need to be a single Markdown file
        // If absolute
          // Check that the URL is reachable, throw error if not

jaredkhan:
Fixes #192 

Adds support for images at relative URLs in notes by resolving relative URLs at the import stage.

### DIFF
--- a/Sources/HaCWebsiteLib/Markdown+relativeURLs.swift
+++ b/Sources/HaCWebsiteLib/Markdown+relativeURLs.swift
@@ -1,0 +1,48 @@
+import Regex
+import Foundation
+
+extension Markdown {
+  /*
+   * Markdown references (links and images) can contain relative URLs
+   * This method will resolve relative URLs relative to a given base
+   * e.g. '[Meow](images/cat.png)' -> '[Meow](https://hackersatcambridge.com/images/cat.png)'
+   */ 
+  mutating func resolveRelativeURLs(relativeTo baseURL: URL) {
+
+    // See https://regexr.com/3jh10 for demo of this regex
+    let referenceTitleRegexString = 
+      "\\[" + // Opening '['
+      "([^\\[\\]]*?)" + // Capture 1: The optional reference title
+      "]" // Closing ']'
+
+    let referenceURLRegexString =
+      "\\(" + // Opening '('
+      "([^)]*?)" + // Capture 2: The URL
+      "\\)" // Closing ')'
+
+    let referenceRegexString =
+      "(!?)" +  // Capture 0: The optional '!' to indicate image references
+      referenceTitleRegexString +
+      referenceURLRegexString
+
+    let referenceRegex = try! Regex(string: referenceRegexString)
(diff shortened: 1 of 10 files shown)


Sources/HaCWebsiteLib/Markdown+relativeURLs.swift
```diff
@@ -0,0 +1,48 @@
+import Regex
+import Foundation
+
+extension Markdown {
+  /*
+   * Markdown references (links and images) can contain relative URLs
+   * This method will resolve relative URLs relative to a given base
+   * e.g. '[Meow](images/cat.png)' -> '[Meow](https://hackersatcambridge.com/images/cat.png)'
+   */ 
+  mutating func resolveRelativeURLs(relativeTo baseURL: URL) {
```
Pinpickle: Why does this have to be mutating? Could it be one of

- Exposed as a string-to-string transformation on markdown sources
- A function returning a new `Markdown` with a different source (`withRelativeURLsResolved`)

Sources/HaCWebsiteLib/Markdown.swift
```diff
@@ -3,7 +3,7 @@ import HaCTML
 
 /// A type for text in Markdown form
 struct Markdown {
-  let raw: String
+  var raw: String
```
Pinpickle: Does this still need to be mutable?

Sources/HaCWebsiteLib/Markdown+relativeURLs.swift
jaredkhan: Went for the latter since doing this as a string-to-string transformation only makes sense in the case of Markdown sources

jaredkhan:
Limitations of this (regex) approach: 
- Doesn't handle commented references
- Doesn't handle nested reference. In this case, will only resolve the innermost reference (e.g. an image within a URL)

Pinpickle:
@jaredkhan if we merge this to keep us going for now, can we document this in code and have an issue to follow up?

jaredkhan:
We sure can @Pinpickle

PR #209, opened 2018-01-23
Use Fluent to automatically run migrations

Pinpickle:
This uses Fluent built-in preparation manager to handle preparations (so they only run once). Previously we were managing this on our own.

I've verified that preparations successfully run, and do not re-run or error when they have already run.

We will need to nuke the DB in prod before this is shipped, but there is no useful data up there.

@CharlieCrisp could you validate that what I am doing is OK?

### DIFF
--- a/Sources/HaCWebsiteLib/Utils/DatabaseUtils.swift
+++ b/Sources/HaCWebsiteLib/Utils/DatabaseUtils.swift
@@ -10,10 +10,6 @@ struct DatabaseURLComponents {
   let database: String
 }
 
-private let preparations = [
-  GeneralEvent.self
-]
-
 extension URL {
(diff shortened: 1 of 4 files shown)


Sources/HaCWebsiteLib/Utils/DatabaseUtils.swift
```diff
@@ -10,6 +10,10 @@ struct DatabaseURLComponents {
   let database: String
 }
 
+private let preparations = [
```
CharlieCrisp: @Pinpickle what is this here for?

Pinpickle: Remnants of a prior implementation, I'll get rid of this

PR #215, opened 2018-01-31
Link new workshop page from homepage 

### LINKED ISSUE #211: Link to workshops from home page and remove `/beta` from route
Once we have all the bits and pieces of workshops ready to go, we can push them live. One way is by integrating them into events, but we also want people to be able to browse all of our workshops on our lovely archive page.

We should be able to go to the workshops page from the home page, and let's drop that `/beta` prefix from the routes while we're at it.

And then we can drop the old workshop data structures and rename `NewWorkshop` -> `Workshop`

martinhartt:
Fixes #211.

- Dropped `/beta` from workshop routes
- Deleted old `Workshop` data structures
- Renamed `NewWorkshop` data structures to `Workshop`
- Added button to workshop in home page

Previously
screen shot 2018-01-31 at 11 52 27

Now
screen shot 2018-01-31 at 11 44 49


Sources/HaCWebsiteLib/WorkshopManager.swift
```diff
@@ -1,62 +1,118 @@
 import Foundation
 import DotEnv
+import Dispatch
+import Yaml
 
 public struct WorkshopManager {
```
Pinpickle: Leaving a note here that Git has awkwardly turned deleting one file and renaming another into a merge of the two files. This is why there's this weird diff going on.

PR #219, opened 2018-02-03
Add a Continous Integration Workshop feature

moosichu:
image

### DIFF
--- a/Sources/HaCWebsiteLib/Views/LandingPage/LandingFeatures/ContinuousIntegrationWorkshop.swift
+++ b/Sources/HaCWebsiteLib/Views/LandingPage/LandingFeatures/ContinuousIntegrationWorkshop.swift
@@ -1,0 +1,21 @@
+import HaCTML
+import Foundation
+
+extension LandingFeatures {
+  static var continuousIntegrationWorkshop: LandingFeature {
+    return EventFeature(
+      eventPeriod: DateInterval(
+        start: Date.from(year: 2018, month: 2, day: 15, hour: 13, minute: 15    ),
+        duration: 1.5 * 60 * 60
+      ),
+      eventLink: "https://www.facebook.com/events/181764052556196/",
+      liveLink: "/workshops/continuous-integration",
+      hero: ImageHero(
+        background: WorkshopManager.workshops["workshop-continuous-integration"]!.promoImageBackground,
+        imagePath: WorkshopManager.workshops["workshop-continuous-integration"]!.promoImageForeground,
+        alternateText: WorkshopManager.workshops["workshop-continuous-integration"]!.title
+      ),
+      textShade: .light
+    )
+  }
+}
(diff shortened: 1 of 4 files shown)


Sources/HaCWebsiteLib/Views/LandingPage/LandingFeatures/ContinuousIntegrationWorkshop.swift
```diff
@@ -0,0 +1,17 @@
...
+extension LandingFeatures {
+  static var continuousIntegrationWorkshop: LandingFeature {
+    return EventFeature(
+      eventPeriod: DateInterval(
+        start: Date.from(year: 2018, month: 2, day: 5, hour: 13, minute: 15    ),
+        duration: 1.5 * 60 * 60
+      ),
+      eventLink: "https://www.facebook.com/events/181764052556196/",
+      liveLink: "/workshops/continuous-integration",
+      hero: WorkshopManager.workshops["workshop-continuous-integration"]!.hero,
```
jaredkhan: Not safe to force unwrap here since it relies on GitHub being up and on the continuous integration workshop data being in a valid format. I'd maybe make LandingFeatures.features be an array of optionals and then have current feature find the first non-optional one with a future expiry date.

moosichu:
@jaredkhan Concerns addressed (hopefully)

Sources/HaCWebsiteLib/Views/LandingPage/LandingFeatures/LandingFeatures.swift
```diff
@@ -20,6 +20,14 @@ enum LandingFeatures {
...
   static var currentFeature: LandingFeature? {
     let currentDate = Date()
-    return features.first { $0.expiryDate > currentDate }
+
+    // We want to return the first non-nil LandingFeature whos expiry date is
+    // after the current date. As the list itself is a list of optionals, we
+    // need to safely unwrap the result at least once as features.first returns
+    // a double optional.
+    return (
+      features.first { $0?.expiryDate ?? currentDate > currentDate }
```
jaredkhan: This is maybe a little hard to read. 
How about: 
`return features.flatMap { $0 }.first{ $0.expiryDate > currentDate }`

Sources/HaCWebsiteLib/Views/LandingPage/LandingFeatures/LandingFeatures.swift
```diff
@@ -23,11 +23,8 @@ enum LandingFeatures {
 
     // We want to return the first non-nil LandingFeature whos expiry date is
```
jaredkhan: Is this comment necessary?

moosichu: I think so, because we want our projects to be accessible as possible. Even if this comment would be a bit redundant to those familiar with Swift, a lot of programmers on this project may not be, so for more complex things a comment could be very helpful.

I would agree, but it's something that I wouldn't have understood unless I knew the problem that needed solving.

PR #220, opened 2018-02-04
Optional fields

jaredkhan:
Makes `thanks` and `further_reading_links` fields of workshop metadata files optional, modelling them as empty lists in Swift when they are omitted from the Yaml file.


jaredkhan:
@moosichu Please look at this when you get a chance

PR #246, opened 2018-02-16
Add event start time to featured event

MasterObvious:
Added time on featured today event

capture

Along with this I added an individualTimeFormatter in DateUtils


MasterObvious:
This should be done now

PR #255, opened 2018-02-17
Workshop refactor

### LINKED ISSUE #250: Create new system that uses workshop-template instead of workshop-example
[workshop-template](https://github.com/hackersatcambridge/workshop-template) is the new version of [workshop-example](https://github.com/hackersatcambridge/workshop-example)

This involves:
- [x] Respecting new file layout (no info/content directory) 
- [x] All markdown files can now use images from the markdown_images directory
- [x] Examples now go in top-level directory so examples button should be updated 

This depends on issue #248 and #249.

jaredkhan:
Fixes #250 

 - Respecting new file layout (no info/content directory)
 - All markdown files can now use images from the markdown_images directory
 - Examples now go in top-level directory so examples button should be updated


Sources/HaCWebsiteLib/Models/Workshop+init.swift
```diff
@@ -53,6 +52,7 @@ extension Workshop {
   }
 
   private static func getMetadata(localPath: String) throws -> Yaml {
+    print (localPath + filePaths.metadata)
```
moosichu: Why is this being printed? If the goal is to log it, why not use our logging system?

jaredkhan: Was just there for debugging. Good catch

PR #259, opened 2018-02-18
Allow slide links to be relative

Pinpickle:
As discussed offline, we want to be able to have a relative ~presentation~ slides link to an html file in the repo which automatically serves the latest version from rawgit.

URLs are relative to the `.hac_workshop` folder.

Tests are added to verify this behaviour.


moosichu:
We can create issue for this though

moosichu:
Hmmm, let me test something.

moosichu:
@Pinpickle Just checked, and my suspisions were right:

https://github.com/moosichu/link-tester/blob/master/README.md

In GitHub markdown, a leading slash always means relative to the root of the repository. Also, I don't know how a leading slash having the meaning of being relative to rawgit.com is any more intuitive.

Pinpickle:
I misunderstood, I thought by "workshop repo", you meant on GitHub instead of rawgit. That sounds very reasonable!

moosichu:
Shall I create an issue for this then so that we can merge this?

Pinpickle:
Go ahead, although it's bigger than these relative URLs (markdown URLs are also affected)

PR #263, opened 2018-02-20
Event Feature text overlay - background added for contrast

### LINKED ISSUE #239: Replace "light" and "dark" text over features with a dark box with light text
Currently, we have text on top of our feature on the home page to display the date. To accommodate for different coloured images, that text can be either "dark" or "light". While a nice design feature, this is a pain to work with. The text should be background-agnostic, eg a dark box and light text.

Here is the code where we create the text on top:

https://github.com/hackersatcambridge/hac-website/blob/master/Sources/HaCWebsiteLib/Views/LandingPage/EventFeature.swift#L29-L44

eliotlim:
+ EventFeature.styl: Added a background and rounded border for the overlay text
- EventFeature.swift: Modified `TextShade` enum to resolve as `EventFeature--text-light`.

Fixes #239.

Before

After

### DIFF
--- a/Sources/HaCWebsiteLib/Views/LandingPage/EventFeature.swift
+++ b/Sources/HaCWebsiteLib/Views/LandingPage/EventFeature.swift
@@ -18,10 +18,8 @@ struct EventFeature: LandingFeature {
     case light
     var styleClass: String {
       switch self {
-        case .light:
+        default:
           return "EventFeature--text-light"
-        case .dark:
-          return "EventFeature--text-dark"
       }
     }
   }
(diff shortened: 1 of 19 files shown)


eliotlim:
Here's a screenshot of the result when a lighter background is used!

Lighter BG Overlay Preview

eliotlim:
I'm inclined to agree. Would it be okay if I removed `TextShade`, or is
there a plan to use it for another visual feature?

CharlieCrisp:
@eliotlim I think it's fine but @jaredkhan or @Pinpickle can you confirm?

jaredkhan:
@eliotlim Yup, `TextShade` should go

eliotlim:
I've removed the `TextShade` enum and updated the relevant LandingFeatures.

PR #270, opened 2018-07-12
Add events editing portal and increase API ability

CharlieCrisp:
This PR: 
- Changes the schema of events to include an id that is unique and used in the url. This is based on the assumption that we currently have no events in our database and can just blitz it rather than using migrations to change the schema.
- Adds new API routes for editing and deleting events in our database. Admin credentials are required.
- Adds a GUI for viewing all events, adding new events, editing and deleting existing events.

(Sorry for the monolithic PR)

# Screenshots:
## Events portal page:
image
## Add new event page:
image
## Edit Event Page:
image
## Delete Event Page:
image
## Individual Event View Page
image

### DIFF
--- a/Sources/HaCWebsiteLib/Controllers/EventApiController.swift
+++ b/Sources/HaCWebsiteLib/Controllers/EventApiController.swift
@@ -25,23 +25,22 @@ struct EventApiController {
             return
           }
         }
-        try saveEvent(json: json)
+        try getEventFromJSON(src: json).save()
         Log.info("Succesfully added event to the database")
-        try response.send("Successfully added your event to the database\n").end()
-      }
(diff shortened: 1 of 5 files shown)


Sources/HaCWebsiteLib/Location.swift
```diff
@@ -1,8 +1,8 @@
 struct Location {
-    let latitude : Double
-    let longitude : Double
-    let address : String?
-    let venue : String?
+    var latitude : Double
```
jaredkhan: Don't see why these should be variables

jaredkhan:
eventApiControllerTests needs updating to reflect these changes.

(Sorry, I know feedback would have been more helpful before you had a full-time job!)

Issue #261, opened 2018-02-18
Use rawgit's CDN when we can


moosichu:
We use https://rawgit.com/ for various things, however, it would make sense to use the CDN version, which we currently aren't.

PR #273, opened 2018-10-18
Change CDN provider for Workshops

### LINKED ISSUE #272: Use new CDN instead of Rawgit
"[Rawgit](https://rawgit.com/) has reached the end of its useful life". It should still be serving things until October 2019 but we may as well change now considering #261 was going to make a change anyway. [JsDelivr](https://www.jsdelivr.com/) is an option. 

Happy to do this.

patricoferris:
Fixes #272  #261 .

Changing the CDN provider from Rawgit (which is now closing down) to JsDelivr

### DIFF
--- a/Sources/HaCWebsiteLib/Models/Workshop+init.swift
+++ b/Sources/HaCWebsiteLib/Models/Workshop+init.swift
@@ -115,8 +115,7 @@ private struct WorkshopBuilder {
   }
 
   private func repoUrl(origin: String, relativePath: String) throws -> URL {
-    let urlString = "\(origin)hackersatcambridge/workshop-\(workshopId)/\(relativePath)"
-
+    let urlString = "\(origin)hackersatcambridge/workshop-\(workshopId)@\(relativePath)"
     if let url = URL(string: urlString) {
       return url
     } else {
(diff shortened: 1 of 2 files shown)


patricoferris:
We'll see if this builds, but currently a lot of the functionality of the website is broken because rawgit is no longer 😢 - I made this PR ages ago to change to `jsDelivr` which *should* fix this...

mukul-rathi:
Let's give this a go!

mukul-rathi (approved):

Sources/HaCWebsiteLib/Models/Workshop+init.swift
```diff
@@ -125,11 +124,11 @@ private struct WorkshopBuilder {
   }
 
   func fileServingUrl(relativePath: String) throws -> URL {
-    return try repoUrl(origin: "https://rawgit.com/", relativePath: "\(commitSha)\(relativePath)")
+    return try repoUrl(origin: "https://cdn.jsdelivr.net/gh/", relativePath: "@\(commitSha)\(relativePath)")
```
eliotlim: What's the @ sign for?
(Forgive my ignorance)

patricoferris: Literally the "@" symbol - it's just the required syntax that `jsDelivr` requires

eliotlim: Ah, that makes sense. I thought it might have been a Swift-specific feature.
String interpolation actually seems pretty cool.

PR #274, opened 2018-12-01
Add redirect to game gig one button edition

patricoferris:
Not a perfect solution, but works for now. Definitely a discussion point in the future about how we're going to manage the site and adding newer content.


mukul-rathi:
Looks good to me!

